PR #239, opened 2015-02-04
Vol 54 again

ginkgomzd:


CRM/Volunteer/Form/Volunteer.php
```diff
@@ -90,16 +143,23 @@ function setDefaultValues() {
...
       $dao->entity_table = CRM_Volunteer_BAO_Project::$_tableName;
-      $dao->entity_id = $this->_project->id;
+      $dao->entity_id = $this->getProject()->id;
       $dao->orderBy('weight asc');
       $dao->find();
       while ($dao->fetch()) {
         $this->_profile_ids[] = $dao->uf_group_id;
       }
     }
+    if (empty($this->_profile_ids)) {
```
ginkgomzd: @GinkgoFJG pretty sure this default ensures backwards compat for existing volunteer projects... However, maybe we still need the upgrade step we discussed to cover upgrading the module and not updating the volunteer project? I suggest merging this and creating a new issue if so.

ginkgomzd:
this also covers VOL-58

CRM/Volunteer/Form/VolunteerSignUp.php
```diff
@@ -285,6 +285,7 @@ function postProcess() {
   function buildCustom() {
     $contactID = CRM_Utils_Array::value('userID', $_SESSION['CiviCRM']);
     $profiles = array();
+    $fieldList = array(); // master field list
```
ginkgomzd: This and the code that uses it below is the fix for the bug reported in the comments of VOL-58 causing improprer multiple-profile display.

PR #246, opened 2015-05-06
VOL-113: Configured profile selector to allow use of profiles which c…

GinkgoFJG:
…ontain custom fields on the Volunteer contact subtype.


ginkgomzd:
@colemanw Do you know if this declaration is needed? Is there such thing as a wildcard? Is empty = all?

more context: https://issues.civicrm.org/jira/browse/VOL-113

colemanw:
FIx looks right to me. The profile code is kind of convoluted and I'm not any more clear than you are about what these options actually do.

PR #247, opened 2015-05-18
VOL-34: Fixes around validation and required fields.

GinkgoFJG:
There are a lot of whitespace changes in this patch; it's not as big as it looks. It might be easier to review this PR commit-by-commit or using the split rather than the unified view.


ginkgomzd:
okay by me to merge this.

PR #251, opened 2015-08-19
VOL-97: Schema changes for CiviVolunteer 2.0 features

GinkgoFJG:


CRM/Volunteer/BAO/Project.php
```diff
@@ -231,11 +239,14 @@ static function retrieveByID($id) {
    * @access public
    */
   public static function dataExists($params) {
```
ginkgomzd: I think this could have a better name. Sounds like a duplicate check function. "validateProjectParams"

GinkgoFJG: Fair enough. That method has been around for a while, though -- pretty sure it was borrowed from some core class -- and it was outside the scope of what I was doing to rename it. Will keep in mind as other refactoring happens.

ginkgomzd:
I leave it to you to take my comment on the dataExists function. Merge at will.

PR #255, opened 2015-08-31
VOL-118: Modify public sign-up form to accommodate multiple projects

GinkgoFJG:


GinkgoFJG:
Reviewed in meatspace by Toby.

PR #259, opened 2015-09-18
VOL-138: set permissions for volunteer_commendation API call

RussellAult:


RussellAult:
I did the change and pull request for the master branch. In hindsight, I'm wondering if I should have changed the 1.4 branch instead. I can re-create the pull request in that branch if that's better.

GinkgoFJG:
This is fine. I'll test in both. Thanks!

PR #264, opened 2015-10-01
Updated country field on the geo search to respect CiviCRM settings (i.e., show only enabled countries).

GinkgoFJG:


GinkgoFJG:
Merging this with the expectation that it will need to be refactored. @TobiasLounsbury, if I need to put a country field on a different screen, the best way to encapsulate the related logic is in a service, no?

PR #265, opened 2015-10-01
Added "Continue" and "Cancel" buttons to complement the "Create/Edit Project" page's "Save" button.

GinkgoFJG:


GinkgoFJG:
@TobiasLounsbury, merged this without review because it seemed uncontroversial. Let me know if you disagree. Note that one of the commits is whitespace-only. Might be quicker to review if you look at only the second diff.

PR #266, opened 2015-10-04
Mixed bag PR: Mostly involves setting defaults and improving themability.

GinkgoFJG:


GinkgoFJG:
@TobiasLounsbury, you might be interested in this in case it overlaps with the work you're doing.

PR #292, opened 2015-11-17
Bugfix/vol 70 multi lang install

JohnFF:
1. The report option creation is now in an api call, fixing https://issues.civicrm.org/jira/browse/VOL-70
2. Added a missing table to the uninstall call. Also made them only call drop if exists.


GinkgoFJG:
Closed in favor of #318.

PR #293, opened 2015-11-20
Remove drupal-only css hack

colemanw:
This had the effect of removing titles from all angular pages (but only on certain drupal themes)
e.g. http://dmaster.demo.civicrm.org/civicrm/a/#/status

There's actually a new angular directive in 4.7 which allows you to set the page title, which IMO is better than hiding it:
https://github.com/civicrm/civicrm-core/blob/master/ang/crmUi.js#L985

### DIFF
--- a/ang/volunteer.css
+++ b/ang/volunteer.css
@@ -43,10 +43,6 @@
   margin-top: 2em;
 }
 
-#tabs-wrapper {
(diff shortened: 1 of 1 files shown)


GinkgoFJG:
Nifty. Thanks.

PR #294, opened 2015-11-30
Bugfix/vol 171 install process fails granular

JohnFF:
Prevent and fix numerous errors on install, uninstall, and reinstalling.


GinkgoFJG:
Closed in favor of #318.

PR #295, opened 2015-11-30
Updated the db populate function to match v 2.0

JohnFF:


GinkgoFJG:
Rejecting this pull request. I no longer remember why, but running the upgrader's install() method from _populateDB() doesn't work. This is the reason that the somewhat redundant code exists in both places.

When I check out your branch and run a unit test -- e.g.,

> cd /usr/local/lib/civicrm-buildkit/build/vol-install/sites/all/modules/civicrm/tools/extensions/civivolunteer/
> git checkout bugfix/update-test-db-installer-to-v2
> cd ../../
> clear && ./scripts/phpunit --include-path extensions/civivolunteer/tests/phpunit/ CRM_Volunteer_BAO_AssignmentTest

... I get a bunch of HTML output on my console with the following language:

> DB Error: no such field

It looks like the install fails after applying your patch.

PR #297, opened 2015-12-01
Vol 154

JohnFF:


GinkgoFJG:
Let's cut down the commit noise by rebasing instead of merging branches.

```
git fetch origin
git checkout VOL-154
git rebase origin/master
```

The above puts the unique commits in VOL-154 after all the commits in master. This is helpful if you're developing against a moving target (i.e., three other PRs get merged from the time you start work to the time you're ready to make your PR). The above also has the benefit of leaving out the totally unnecessary merge commit ded9894.

There's a caveat. If you're rebasing locally, you're changing history. When you try to push to your remote, you may find that it's rejected unless you `git push -f JohnFF`. Pushing with the -f (force) flag should be avoided when working on a shared repository, but in this case the push is happening against a temporary branch in a fork that no one else is working on, so it's okay.

Here's a [good introduction](http://nathanleclaire.com/blog/2014/09/14/dont-be-scared-of-git-rebase/) to rebase and push -f.

JohnFF:
Hi Frank,
Latest version here. I have still to do the rebase. It has changed significantly.
In order for unsetting the campaign to work, it can't be looked up from the project. I'll take a second look at it tomorrow, but currently it's being passed in.

PR #298, opened 2015-12-02
Vol 167 owner

TobiasLounsbury:


GinkgoFJG:
I think I found a bug. If I edit an existing project and change the values of most fields, I get the expected save-and-redirect behavior when I click "Save and Done." If I edit the relationships, however, nothing happens when I click the button.

PR #306, opened 2015-12-20
Fixed regression resulting from the decision to store profile module_data as JSON.

GinkgoFJG:
The resulting behavior was that additional volunteers (aka group registrations) could sign up to any
volunteer project, even if the profiles were configured to allow individual sign-up only.


GinkgoFJG:
@TobiasLounsbury, could you please review this PR? @adam-edison discovered a regression while writing doc; this PR should fix it.

Because of how the perms are set up, I'll still have to merge it myself, but at least you can provide peer-review. Thanks!

PR #308, opened 2015-12-30
VOL-174: When creating new projects, users lacking access to view the default contacts for the project relationships can't submit the form.

GinkgoFJG:
Addresses the problem by:
* removing the crmEntity widgets when they are harmful rather than helpful
* adding form validation to ensure that all the core project relationships are set

### DIFF
--- a/ang/volunteer/Project.js
+++ b/ang/volunteer/Project.js
@@ -55,6 +55,12 @@
           },
           profile_status: function(crmProfiles) {
             return crmProfiles.load();
+          },
+          // VOL-174
+          userCanGetContactList: function (crmApi) {
+            return crmApi('Contact', 'getlist').then(function(result) {
+              return (result.count > 1);
+            });
           }
         }
       });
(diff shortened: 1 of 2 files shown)


ang/volunteer/Project.js
```diff
@@ -218,11 +224,37 @@
...
+      var isValid = true;
+
+      var requiredRelationshipTypes = ['volunteer_beneficiary', 'volunteer_manager', 'volunteer_owner'];
+
+      _.each(requiredRelationshipTypes, function(value) {
+        var thisRelType = _.find(supporting_data.values.relationship_types, function(relType) {
+          return (relType.name === value);
+        });
+
+        if (!_.isArray(relationships[thisRelType.value]) || _.isEmpty(relationships[thisRelType.value])) {
```
GinkgoFJG: Assuming that `thisRelType` will have a `value` property presents a small but acceptable risk.

The required relationship types are the machine names of the option group values installed by the extension, which are reserved. Users cannot delete them, disable them, or change the machine name without mucking around in the database. `thisRelType` should always have a `value` property in normal situations.

I'm not sure what an appropriate error message would be for the user in the case someone had mucked up the DB, so I think it's probably okay to leave this as is.

PR #309, opened 2016-01-04
Next batch of changes to documentation.

adam-edison:


GinkgoFJG:
Hi, @adam-edison,

Thanks for the pull request. Here's a little feedback:
* The main document was auto-generated by GitHub. The editing instructions probably don't apply (at least not entirely) to most contributors -- only to the project administrator, and only if they are using the GitHub editor. I would not assume that this page will look this way for very long; someone else is working on content to replace this page. Perhaps it would be more useful to add some navigation to the header so that it appears on every page?
* Your pull request is very good at describing what is on the page, but I think it could be improved by describing what those things mean. For example, is it clear to you what the "Associated Entity" column is for? Will it be clear to the reader?

adam-edison:
I've commented out the main document changes, and added a navigation button to the header instead. So, that should solve that problem for the time being.

I've read through the Wiki, Blog, and even the developer's notes on Confluence, and even though I see some general language about Volunteer Projects being referred to as Entities, I cant figure out what that _Associated Entity_ field is for. I would appreciate some enlightenment on that.

PR #311, opened 2016-01-13
Added a (non-configurable) date sorting to search results.

GinkgoFJG:


CRM/Volunteer/BAO/NeedSearch.php
```diff
@@ -85,6 +85,7 @@ public function search() {
     }
 
     $this->getSearchResultsProjectData();
+    $this->sortSearchResults();
```
GinkgoFJG: Yeah, that's fine. Your change makes it more flexible because theoretically there could be a config to specify how we want results sorted. I couldn't decide whether it was writing YAGNI, so I went for what was most expedient.

PR #313, opened 2016-01-15
Port fetching of volunteer custom fields from JS to PHP.

GinkgoFJG:
Could @TobiasLounsbury or @ginkgomzd take a look at this, please?


TobiasLounsbury:
This seems like a good approach, With the exception of the one line comment It's ready to merge.

PR #314, opened 2016-01-15
Replace menu-level access control with fine-grained context-aware access control

GinkgoFJG:
Could @TobiasLounsbury or @ginkgomzd review this, please?

Thinking about this some more, I wonder if I didn't go far enough. Perhaps we should skip checking perms at the menu entry level altogether and do like we do for [the public sign up form](https://github.com/civicrm/org.civicrm.volunteer/blob/v4.6-2.0.beta2/CRM/Volunteer/Form/VolunteerSignUp.php#L132-L138).

Benefits:
* We can use CRM_Volunteer_Permission::checkProjectPerms() to ensure the user has permission to access this particular log.
* Consistency with regard to how we deal with Joomla instances' poor handling of extension-based permissions.

Quack, quack.


GinkgoFJG:
@TobiasLounsbury, please have a look at the new commits. If you approve I'll rebase and merge. Thanks!

GinkgoFJG:
Peer-reviewed by @TobiasLounsbury.

PR #318, opened 2016-01-20
Improve install/uninstall code

GinkgoFJG:
Refactors installation code to address VOL-171. Refactors other contributed code, mostly around uninstallation.


CRM/Volunteer/Upgrader.php
```diff
@@ -68,27 +68,22 @@ public function install() {
...
       ));
-      $optionGroupId = $optionGroup['id'];
     } catch (Exception $e) {
-      // if an exception is thrown, most likely the option group already exists,
-      // in which case we'll just use that one
-      $optionGroupId = civicrm_api3('OptionGroup', 'getvalue', array(
-        'name' => CRM_Volunteer_BAO_ProjectContact::RELATIONSHIP_OPTION_GROUP,
-        'return' => 'id',
-      ));
+      $msg = 'Exception thrown in ' . __METHOD__ . '. Likely the option group alreadt exists.';
```
ginkgomzd: typoe

CRM/Volunteer/Upgrader.php
```diff
@@ -116,7 +111,12 @@ public function installProjectRelationships() {
     );
 
     foreach ($options as $opt) {
-      civicrm_api3('OptionValue', 'create', $optionDefaults + $opt);
+      $getOptionValues = civicrm_api3('OptionValue', 'get', $optionDefaults + $opt);
```
ginkgomzd: seems wrong that defaults would not be overwritten, as would be the case if using array_merge instead.

ginkgomzd:
Looks good to merge after fixing the typo in message to users and validation of the array operation.

CRM/Volunteer/Upgrader.php
GinkgoFJG: Meh. We probably do want those values to be set that way all the time, and it's highly unlikely the values would come from an API call since we are installing new options, but I agree that it's intuitive that defaults should be mutable, so I've pushed another commit.

PR #340, opened 2016-02-23
Log fact that associated entity is missing rather than allow the app to crash.

GinkgoFJG:
If a volunteer project is associated with an event and the event is later deleted,
CiviVolunteer will log this instead of allow the application to crash and burn when
details are requested about the associated entity.


GinkgoFJG:
FYI @TobiasLounsbury and @axon-obriend.

TobiasLounsbury:
This is good to go

PR #343, opened 2016-02-24
VOL-175: Fixed rendering of menus for moving/copying volunteers on the Assign UI.

GinkgoFJG:
Sidestepped race condition by deferring DOM lookup until menuItemTemplate() is called.


GinkgoFJG:
Reviewed by @TobiasLounsbury.

PR #347, opened 2016-03-22
 s/Multi-static/Multi-Static/g 

k054:
lower 's' was causing errors durin intallation


GinkgoFJG:
To provide a little more context:

@k054 reported on IRC:

> I have tried installing civi volunteer twice on a fresh civi install and I got "Multi-select' is not a valid option for field html_type" every time

GinkgoFJG:
@k054, thanks for the fix. Your contribution is most welcome, and I hope you'll continue to contribute to the project.

Did patching this allow you to install the extension?

k054:
it did, out of the box from the Civi extension manager.

However the Angular Profiles (org.civicrm.angularprofiles) it not been shown in the extensions list and I needed to get it from here: https://github.com/ginkgostreet/org.civicrm.angularprofiles

thanks for the effort.

PR #349, opened 2016-04-01
Missingeventid

alifrumin:
Fixes bug when registering to volunteer for one event and that event does not have an entity id

### DIFF
--- a/CRM/Volunteer/Form/VolunteerSignUp.php
+++ b/CRM/Volunteer/Form/VolunteerSignUp.php
@@ -682,7 +682,8 @@ class CRM_Volunteer_Form_VolunteerSignUp extends CRM_Core_Form {
         if (count($this->_projects) === 1) {
           $project = current($this->_projects);
           if (empty($project['entity_id'])) {
-            continue;
+            $path = 'civicrm/vol/';
+            $fragment = '/volunteer/opportunities';
           }
           $eventId = $project['entity_id'];
           $path = 'civicrm/event/info';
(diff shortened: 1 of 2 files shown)


GinkgoFJG:
Hi, @alifrumin! Thanks for contributing to CiviVolunteer! Sorry for the delayed response -- it was a busy week.

Can you say a little bit more about the problem you encountered that let you to make these changes? From the looks of the patch, I don't think this will fix them. I'll comment in more detail inline with the code.

Thanks!

CRM/Volunteer/Form/VolunteerSignUp.php
```diff
@@ -681,6 +681,10 @@ protected function setDestination() {
         // to that event form; otherwise, default to the vol opps page.
         if (count($this->_projects) === 1) {
           $project = current($this->_projects);
+          if (empty($project['entity_id'])) {
```
GinkgoFJG: I'm a little surprised that you're needing to check for the entity id since you're inside the 'event' case. Unless there's a bug somewhere else -- or you're generating your own links somehow -- the `dest` parameter should only be set to 'event' if you've arrived at the sign up form by clicking the "Volunteer Now" button from an Event Info page.

CRM/Volunteer/Form/VolunteerSignUp.php
```diff
@@ -681,6 +681,10 @@ protected function setDestination() {
...
+            $path = 'civicrm/vol/';
+            $fragment = '/volunteer/opportunities';
+          }
           $eventId = $project['entity_id'];
           $path = 'civicrm/event/info';
```
GinkgoFJG: The `$path` you set above (line 685) gets overridden here.

alifrumin:
Hey Frank,

You are right. Sorry for the confusion and thank you for your feedback. 

Best

PR #350, opened 2016-04-08
Vol 189 event info button

TobiasLounsbury:


GinkgoFJG:
@TobiasLounsbury, I think this is an inadequate check for "open needs."

(As an aside, there's no need to wrap _get_open_needs(). It's a delegate of PHP's magic get, so requesting the property (i.e., `$project->open_needs`) returns the result of the private method.)

More importantly, though, there is a definition for an open need, and your use case actually requires more than just the open needs. According to the code doc, "open" means:
* that the number of volunteer assignments associated with the need is fewer than quantity specified for the need
* that the need's start time or end time is in the future
* that the need is active
* that the need is visible
* that the need has a start_time (i.e., is not flexible)

Your check also needs to check for flexible needs, i.e., those needs which allow a user to sign up without specifying a preference for time or role.

It might be more reliable to use the same query that the search form itself uses:

```php
civicrm_api3('VolunteerNeed', 'getsearchresult', array(
  'project' => $projectId,
));
```

PR #351, opened 2016-04-08
Added URL params to hide search pane with varying levels of force

TobiasLounsbury:


ang/volunteer/VolOppsCtrl.html
```diff
@@ -16,7 +16,12 @@
   </div>
 
   <!-- Search Form -->
-  <div class="crm-vol-half-responsive crm-vol-opp-search">
+  <div class="crm-vol-half-responsive crm-vol-opp-search" ng-show="allowShowSearch">
```
GinkgoFJG: The wrapper for the button has the same class name as the wrapper for search form. Copy/paste error?

ang/volunteer/VolOppsCtrl.js
```diff
@@ -31,6 +31,29 @@
     // on page load, search based on the URL params
     volOppSearch.search();
 
+    //VOL-190: Allow hiding of search pane based on url param
+    if ($route.current.params.hasOwnProperty('hideSearch')) {
```
GinkgoFJG: What if we simplified this a bit?

```javascript
$scope.hideSearch = false;
$scope.allowShowSearch = true;

if ($route.current.params.hasOwnProperty('hideSearch')) {
  if ($route.current.params.hideSearch === "1" ||
    $route.current.params.hideSearch === "always"
  ) {
    $scope.hideSearch = true;    
  }
  if ($route.current.params.hideSearch === "always") {
    $scope.allowShowSearch = false;
  }
}
```

This is a slight deviation from your spec, but really we want the default to be that the search form is displayed. To deviate from the default, you should be able to get the value right: '1' or 'always.' No need to be so generous to lazy developers, I say.

ang/volunteer/VolOppsCtrl.html
GinkgoFJG: To ensure we don't show the button when the form is already displayed, how about we do:

`ng-show="allowShowSearch && hideSearch"`?

(My simplification of the logic in the controller requires this change.)

PR #355, opened 2016-05-11
VOL-199: Updated setDestination() to reset pointer of project list be…

TobiasLounsbury:
…fore attempting to pull data.


GinkgoFJG:
@TobiasLounsbury, @ginkgomzd: I'm not sure if this bugfix is a good candidate or not, but in general I'd like to see unit tests committed along with bugfixes to avoid regressions later.

ginkgomzd:
I reviewed it pretty carefully. As a point of procedure, not all bug fixes come with changes to tests. 
But, point taken.

TobiasLounsbury:
Given that our test coverage is less than perfect, it might be a good idea to make a point of adding tests for any code we modify when it doesn't have an existing test,(or updating if the profile of that test has changed) as a way of incrementally increasing our code coverage without needing to spend an arduous amount of time in a lump sum.

PR #356, opened 2016-05-18
Added button classes to three submit buttons on edit project screen

TobiasLounsbury:


GinkgoFJG:
@TobiasLounsbury, you added a bunch of classes here, and I can see patterns in the naming conventions, and I can tell you that we have not been so consistent with class names in the past. I wonder if a style doc of sorts would be helpful to guide naming conventions going forward...

TobiasLounsbury:
I think a style guide would be a great idea. I added classes that seemed to make sense to me in terms of how I would want classes added if I was creating a custom theme for Volunteer. And I think that is a good practice going forward, but I can't expect everyone to think the same way, so perhaps we should tag-team on a style guide?

PR #358, opened 2016-05-19
Vol 200

TobiasLounsbury:


ang/volunteer/Project.js
```diff
@@ -75,6 +73,8 @@
 
     var relationships = {};
     if(project.id == 0) {
+      project = supporting_data.values.defaults;
```
GinkgoFJG: Overwriting the existing object seems bad, even though at this point there isn't much in it. How about instead:

```javascript

project = _.extend(supporting_data.values.defaults, project);
```

This way, we use the defaults as the baseline object and anything that has already been defined in the project object (e.g., the ID) overrides the defaults. This has no effect now, but may prevent future head-scratchers as the code base matures.

settings/volunteer.setting.php
```diff
@@ -13,4 +13,90 @@
     'description' => 'Fields which are configured to use the slider widget',
     'help_text' => 'Which fields should use the slider widget?',
   ),
+
+  'volunteer_default_profiles_individual' => array(
+    'group_name' => 'CiviVolunteer Configurations',
+    'group' => 'volunteer_defaults',
+    'name' => 'volunteer_default_profiles_individual',
+    'type' => 'String',
+    'default' => '',
```
GinkgoFJG: You've changed the behavior here. If the user takes no action, the default individual profile should be `volunteer_sign_up`, as was previously hardcoded into `civicrm_api3_volunteer_util_getsupportingdata`.

volunteer.php
```diff
@@ -121,19 +121,33 @@ function volunteer_civicrm_navigationMenu(&$params) {
         ),
         $newNavId + 5 => array(
           'attributes' => array(
+            'label' => ts('Configure Default Project Settings', array('domain' => 'org.civicrm.volunteer')),
+            'name' => 'volunteer_config_defaults',
+            'url' => 'civicrm/volunteer/defaults',
```
GinkgoFJG: If we add defaults for other entities, such as Needs, would those appear here as well? If not, the URL should specify "project."

volunteer.php
```diff
@@ -650,4 +664,62 @@ function volunteer_civicrm_fieldOptions($entity, $field, &$options, $params) {
...
+ * This function fetches the defaults from civicrm settings
+ * And puts them into the appropriate data format to return
+ * to the angular front-end
+ *
+ * @return array
+ * @throws CiviCRM_API3_Exception
+ */
+function volunteer_composeDefaultArray() {
+  $defaults = array();
+  $settings = CRM_Core_BAO_Setting::getItem("volunteer_defaults");
```
GinkgoFJG: The behavior you may be expecting was [deprecated in 4.7](https://github.com/civicrm/civicrm-core/commit/3a84c0ab9022feeed11adf8f9023fd5f2d4459ba#diff-eaa4bbbf67c4d03afd071da3748383abR241).

Totten provided some background on the API list:

> [A]s soon as we reworked the setting layer in 4.7, the runtime for tests dropped by ~80% (3hr vs 40min). As part of the new regime, settings are loaded by exactly two queries during bootstrap. Pseudocode:
>
>  SELECT name, value FROM civicrm_setting WHERE domain_id = %1;
>  SELECT value FROM civicrm_cache WHERE path = ‘setting-defaults’;
>
> It doesn’t matter how many times you call the setting functions — it’s still just the two original queries. In theory, the balance could change in the future (eg if the settings size grow by 100x), but the current numbers don’t support those theories. (As a notable reference point… D6+D7 have similar performance profile for variable_get(). It only changed in D8 after they needed to account for massive growth in contrib and after they expanded the scope of ‘config’ to include things which had previously been distinct tables.)
>
> So… the theoretical concern does exist, and it’s reasonable to have some kind of get-by-prefix or get-by-group function in anticipation. For now, it can just be a simple filter around Civi::settings()->all();  but in the future it could be something more complicated.

Alternately, you could fetch them one by one:

```php
$foo = CRM_Core_BAO_Setting::getItem('myGroup', 'foo');
```

volunteer.php
```diff
@@ -650,4 +664,62 @@ function volunteer_civicrm_fieldOptions($entity, $field, &$options, $params) {
...
+
+  $defaults['is_active'] = (array_key_exists("volunteer_default_is_active", $settings)) ? $settings['volunteer_default_is_active'] : 1;
```
GinkgoFJG: More succinctly:

```php
$defaults['is_active'] = CRM_Utils_Array::value('volunteer_is_active', $settings, 1);
```

volunteer.php
```diff
@@ -650,4 +664,62 @@ function volunteer_civicrm_fieldOptions($entity, $field, &$options, $params) {
...
+  $defaults['campaign_id'] = (array_key_exists("volunteer_default_campaign", $settings)) ? $settings['volunteer_default_campaign'] : '';
+  $defaults['loc_block_id'] = (array_key_exists("volunteer_default_locblock", $settings)) ? $settings['volunteer_default_locblock'] : '';
+
+  $coreDefaultProfile = array(
```
GinkgoFJG: It feels like too much is going on with the profiles in this function. I'd argue that the defaults you're setting in `$coreDefaultProfile` belong in volunteer.setting.php.

Also, the foreach loop below overrides the database. Why not get the weight, for instance, from the `civicrm_uf_join` table instead of forcing one? Likewise, if in the future we load more into `module_data`, this code will have to be rewritten.

Finally, I haven't reviewed the PR in its entirety yet, but I haven't seen the reason for having three separate settings for the different uses of the profile (individual, group, both). If we add a new use, will we then need a new setting? Why not one ring to rule them all? Pseudocode:

```
var volunteerProjectDefaultProfile = [
  {
    use: 'primary',
    id: 25
  },
  {
    use: 'primary',
    id: 19
  },
  {
    use: 'both',
    id: 77
  }
];
```

... or ...

```
var volunteerProjectDefaultProfile = {
  primary: [1, 40, 12],
  additional: []
  both: [3]
};
```

volunteer.php
```diff
@@ -650,4 +664,62 @@ function volunteer_civicrm_fieldOptions($entity, $field, &$options, $params) {
...
+        $profile['module_data']['audience'] = $audience;
+        $profiles[] = $profile;
+      }
+    }
+  }
+
+  $defaults['profiles'] = $profiles;
+
+
+  //Todo: Handle Contacts
```
GinkgoFJG: I think what you're referring to as _Contacts_ is referred to in other places as _Project Relationships_. This is a good idea -- might want to create a JIRA ticket so it doesn't get forgotten.

volunteer.php
```diff
@@ -650,4 +664,62 @@ function volunteer_civicrm_fieldOptions($entity, $field, &$options, $params) {
...
+
+/**
+ * This function fetches the defaults from civicrm settings
+ * And puts them into the appropriate data format to return
+ * to the angular front-end
+ *
+ * @return array
+ * @throws CiviCRM_API3_Exception
+ */
+function volunteer_composeDefaultArray() {
```
GinkgoFJG: Presently this function is used only by civicrm_api3_volunteer_util_getsupportingdata(). Unless you anticipate other uses for this function, I'd move it to api/v3/VolunteerUtil.php and prefix the name with an underscore to indicate it shouldn't be called publicly.

volunteer.php
GinkgoFJG: Since the settings group no longer gains you the convenience factor during retrieval, why not use the same settings group as the other volunteer settings for consistency?

CRM/Volunteer/Form/Defaults.php
```diff
@@ -0,0 +1,154 @@
...
+      ),
+    ));
+    // export form elements
+    $this->assign('elementNames', $this->getRenderableElementNames());
+    parent::buildQuickForm();
+  }
+
+
+  function setDefaultValues() {
+    $defaults = CRM_Core_BAO_Setting::getItem("volunteer_defaults");
```
GinkgoFJG: Flagging another use of getItem that works differently in 4.7.

CRM/Volunteer/Form/Defaults.php
```diff
@@ -0,0 +1,154 @@
...
+  function postProcess() {
+    $values = $this->exportValues();
+
+    CRM_Core_BAO_Setting::setItem($values['volunteer_default_profiles_individual'],"volunteer_defaults", "volunteer_default_profiles_individual");
+    CRM_Core_BAO_Setting::setItem($values['volunteer_default_profiles_group'],"volunteer_defaults", "volunteer_default_profiles_group");
+    CRM_Core_BAO_Setting::setItem($values['volunteer_default_profiles_both'],"volunteer_defaults", "volunteer_default_profiles_both");
+    CRM_Core_BAO_Setting::setItem($values['volunteer_default_campaign'],"volunteer_defaults", "volunteer_default_campaign");
+    CRM_Core_BAO_Setting::setItem($values['volunteer_default_locblock'],"volunteer_defaults", "volunteer_default_locblock");
+
+    $isActive = array_key_exists("volunteer_default_is_active", $values) ? $values['volunteer_default_is_active'] : 0;
```
GinkgoFJG: Use array util instead.

CRM/Volunteer/Form/Defaults.php
```diff
@@ -0,0 +1,154 @@
...
+    CRM_Core_BAO_Setting::setItem($isActive, "volunteer_defaults", "volunteer_default_is_active");
+
+    CRM_Core_BAO_Setting::setItem($values['volunteer_default_contacts'],"volunteer_defaults", "volunteer_default_contacts");
+
+    CRM_Core_Session::setStatus(ts("Changed Saved"), "Saved", "success");
```
GinkgoFJG: Typo: Changes saved.

CRM/Volunteer/Hook.php
```diff
@@ -0,0 +1,13 @@
+<?php
+class CRM_Volunteer_Hook
+{
+  static $_nullObject = NULL;
+
+  static function runDefaultSettingsHook(&$defaults) {
+    return CRM_Utils_Hook::singleton()->invoke(1, $defaults, self::$_nullObject,
+      self::$_nullObject, self::$_nullObject, self::$_nullObject, self::$_nullObject,
+      'volunteer_ProjectDefaultSettings'
```
GinkgoFJG: By convention, I think this hook should be called _civicrm_volunteer_ProjectDefaultSettings_.

CRM/Volunteer/Hook.php
```diff
@@ -0,0 +1,13 @@
+<?php
+class CRM_Volunteer_Hook
+{
+  static $_nullObject = NULL;
+
+  static function runDefaultSettingsHook(&$defaults) {
```
GinkgoFJG: By convention this method take the name of the hook, minus the prefix. I think that means we'd call it _volunteer_projectDefaultSettings_ or _projectDefaultSettings_. I think I prefer the latter, since _volunteer_ is already in the class name.

GinkgoFJG: Code doc, please. See [buildForm hook](https://github.com/civicrm/civicrm-core/blob/ef587f9c568bb8e0e310b910c905d61fb507da54/CRM/Utils/Hook.php#L368-L379), for example.

CRM/Volunteer/Hook.php
```diff
@@ -0,0 +1,13 @@
+<?php
+class CRM_Volunteer_Hook
+{
+  static $_nullObject = NULL;
```
GinkgoFJG: Don't think you need this; I think you can use CRM_Utils_Hook::$_nullObject instead.

api/v3/VolunteerUtil.php
```diff
@@ -143,13 +143,17 @@ function civicrm_api3_volunteer_util_getsupportingdata($params) {
...
+    $defaults = volunteer_composeDefaultArray();
+
+    //StopGap because the interface for contacts didn't fit into scope
+    if(!array_key_exists("relationships", $defaults)) {
+     $defaults['relationships'] = _volunteerGetProjectRelationshipDefaults();
+    }
+    //Allow other extensions to modify the defaults
+    CRM_Volunteer_Hook::runDefaultSettingsHook($defaults);
+
+    $results['defaults'] = $defaults;
```
GinkgoFJG: So, I'm having misgivings about the use of the _defaults_ namespace -- not just here but in general (e.g., setting names). It's a little broad.

In the case of this API, is there a chance we'd return different defaults, or defaults for a different object altogether, depending on the value of `$params['controller']`? How do the answers to those questions inform our opinion of the appropriateness of calling the result `$results['defaults']`?

In the case of the settings themselves, mightn't we in the future decide we also need configurable defaults for needs? In that case, wouldn't a naming scheme like `volunteer_default_is_active` introduce the likelihood of naming collisions... or at least confusion?

xml/Menu/Volunteer.xml
```diff
@@ -103,4 +103,10 @@
     <access_arguments>edit own volunteer projects</access_arguments>
     <is_public>true</is_public>
   </item>
+  <item>
+    <path>civicrm/volunteer/defaults</path>
```
GinkgoFJG: See [previous comment](https://github.com/civicrm/org.civicrm.volunteer/pull/358/files#diff-0bf5c137ed2d78e85f555b5dfb9f3003R126).

volunteer.php
TobiasLounsbury: It is not specific to the API, and I do foresee it being used elsewhere in future. It really should go in the ProjectUtil BAO, which doesn't exist. Perhaps I should move it to the Project BAO, but I don't think it belongs in the API

GinkgoFJG: Project BAO works for me. Thanks!

volunteer.php
TobiasLounsbury: I have collapsed the three profile types into a single setting and added some audience type definitions in the Project BAO.
The loop you reference doesn't override any database entry. The data that is being looped over in this function is the defaults settings, which doesn't contain a weight. Only an audience type and a profile ID. The loop is composing the needed data structure for the front end to use. And as weight is a required field, I'm adding it. It seems a bit arbitrary but it should correlate to the order in which items are visible in both the Defaults, and Project Management UI.

CRM/Volunteer/Form/Defaults.php
```diff
@@ -18,32 +18,16 @@ function buildQuickForm() {
...
-      ts('Default Profile(s) for Both Individual and Group Registration'),
-      $profileList,
-      false, // is required,
-      array("placeholder" => ts("-- No Default profiles --"), "multiple" => "multiple", "class" => "crm-select2")
-    );
+    foreach(CRM_Volunteer_BAO_Project::getProjectProfileAudienceTypes() as $audience) {
+      $this->add(
+        'select',
+        'volunteer_default_profiles_' . $audience['type'],
+        ts($audience['description']),
```
GinkgoFJG: This doesn't work. The original strings need to be ts'ed, not the variables that represent them.

GinkgoFJG: Also the ts should reference the extension name as the domain.

CRM/Volunteer/Form/Defaults.php
```diff
@@ -18,32 +18,16 @@ function buildQuickForm() {
...
+        $profileList,
+        false, // is required,
+        array("placeholder" => ts("-- No Default profiles --"), "multiple" => "multiple", "class" => "crm-select2")
```
GinkgoFJG: See earlier notes on ts.

CRM/Volunteer/Form/Defaults.php
```diff
@@ -114,18 +106,29 @@ function validate() {
...
-    CRM_Core_BAO_Setting::setItem($values['volunteer_default_campaign'],"volunteer_defaults", "volunteer_default_campaign");
-    CRM_Core_BAO_Setting::setItem($values['volunteer_default_locblock'],"volunteer_defaults", "volunteer_default_locblock");
 
-    $isActive = array_key_exists("volunteer_default_is_active", $values) ? $values['volunteer_default_is_active'] : 0;
-    CRM_Core_BAO_Setting::setItem($isActive, "volunteer_defaults", "volunteer_default_is_active");
+    //Compose the profiles before we save tem.
+    $profiles = array();
+
+    foreach(CRM_Volunteer_BAO_Project::getProjectProfileAudienceTypes() as $audience) {
+      $profiles[$audience['type']] = CRM_Utils_Array::value('volunteer_default_profiles_' . $audience['type'], $values);
```
GinkgoFJG: Do we need to do any type checking here? Seems like there's a chance we'll get an empty string or some other falsey value we want to represent as an empty array...

ang/volunteer/Project.html
```diff
@@ -150,9 +150,7 @@
           <div class="crm-volunteer-profile-use-type">
             <label>{{ts("Use For:")}}</label>
             <select ng-model="profile.module_data.audience" crm-ui-select="{}" ng-change="validateProfileSelections()">
-              <option value="primary">{{ts("Individual Registration")}}</option>
-              <option value="additional">{{ts("Group Registration")}}</option>
-              <option value="both">{{ts("Both")}}</option>
+              <option ng-repeat="audience in supporting_data.profile_audience_types" value="{{audience.type}}">{{ts(audience.label)}}</option>
```
GinkgoFJG: No need for the ts here.

CRM/Volunteer/Form/Defaults.php
TobiasLounsbury: I don't think it matters. I've tested it with single, multiple and null values and they all work fine. This only stores the setting, and when we parse the setting later we are looking for values, and only populate based on the presence of those values.

volunteer.php
GinkgoFJG: Makes sense. I was probably conflating properties of the profile with those of the join...

volunteer.php
GinkgoFJG: See [VOL-202](https://issues.civicrm.org/jira/browse/VOL-202).

PR #361, opened 2016-05-24
Changed string  'Open until filled' to 'Open-Ended' on Define Opportun…

TobiasLounsbury:
…e Manage Projects screen


GinkgoFJG:
@TobiasLounsbury, I tried to tame the number of classes a bit. I think this is better.

PR #363, opened 2016-05-26
Vol 203 campaign filters

TobiasLounsbury:


CRM/Volunteer/Form/Defaults.php
```diff
@@ -156,8 +200,9 @@ function getRenderableElementNames() {
 
     foreach ($this->_elements as $element) {
       $name = $element->getName();
-      $entity = preg_replace("/volunteer_(.*)_default_.*/", "$1", $name);
-      $group = ts("Default " . ucfirst($entity) . " Settings", array('domain' => 'org.civicrm.volunteer'));
+      $entity = preg_replace("/volunteer_([a-zA-Z0-9]*)_.*/", "$1", $name);
+      $includeDefault = strpos($name, "default") ? "Default " : "";
+      $group = ts($includeDefault . ucfirst($entity) . " Settings", array('domain' => 'org.civicrm.volunteer'));
```
GinkgoFJG: The magic has gotten too magical :-) This can't be translated because of how it's generated.

We should really refactor this form to be built based on meta data ([example](https://github.com/eileenmcnaughton/nz.co.fuzion.civixero/blob/master/CRM/Civixero/Form/XeroSettings.php)), but I'm going to leave this for now, with the exception of taking out the ts() function call.

TobiasLounsbury: I cleaned this up and added metadata support in the associated issue VOL-204

PR #366, opened 2016-05-27
Added logic to allow country list to be empty

TobiasLounsbury:


GinkgoFJG:
It looks like this is a bug fix. Could you please create a JIRA issue saying more about how the bug manifests and the conditions needed for it to arise so we can create a little more context/history for this change? Thanks.

TobiasLounsbury:
I've created a Jira issue https://issues.civicrm.org/jira/browse/VOL-205 to outline this issue, and how to reproduce.

GinkgoFJG:
Closing in favor of #374. I independently stumbled on the same bug and wrote a patch that I prefer. I have not run the code in #366, but if I'm reading it right, in the case of an undefined country limit the code would return all countries, which doesn't seem right. The other patch will cause the API to always return at least the default country (which, through the UI at least, cannot be unset); this seems like a more reasonable fallback.

GinkgoFJG:
Actually, on another read of both your issue and the relevant UI help text, I see that an empty country limit list is supposed to be treated as ALL countries, in which case your PR is probably more appropriate than mine. Reopening and re-investigating.

PR #373, opened 2016-06-05
VOL-196: update schema metadata for compatibility with core API changes

GinkgoFJG:
* [VOL-196: API Basic Get ignores entity definition](https://issues.civicrm.org/jira/browse/VOL-196)


GinkgoFJG:
Reviewed at the code sprint by Heather Killam.

PR #375, opened 2016-06-06
VOL-207 Replace image with text for delete.

hkillam:
* [VOL-207: Broken "delete" icon in Define Volunteer Opportunities UI](https://issues.civicrm.org/jira/browse/VOL-207)


GinkgoFJG:
Thanks for the fix!

PR #383, opened 2016-06-07
Migrate settings access from BAO to API

GinkgoFJG:
This PR also includes a fix for a minor bug in which the default value for a setting couldn't be read because it was formatted improperly. It's minor because:
- the setting in question still hasn't been part of a formal release
- user-provided values for the setting were being formatted properly and could be retrieved and used as expected


GinkgoFJG:
Reviewed in meatspace by Eileen.

PR #392, opened 2016-06-17
Refactor loc block changes

GinkgoFJG:
@TobiasLounsbury, I made this change so that the code that was previously triggered by locBlockChanged (namely setting of the default country), could be triggered were the location block ID to be changed before the form is rendered (i.e., by crmHook).

### DIFF
--- a/ang/volunteer/Project.js
+++ b/ang/volunteer/Project.js
@@ -61,6 +61,24 @@
 
 
   angular.module('volunteer').controller('VolunteerProject', function($scope, $location, $q, $route, crmApi, crmUiAlert, crmUiHelp, countries, project, profile_status, campaigns, relationship_data, supporting_data, location_blocks, volBackbone) {
+
+    /**
+     * We use custom "dirty" logic rather than rely on Angular's native
+     * functionality because we need to make a separate API call to
+     * create/update the locBlock object (a distinct entity from the project)
+     * if any of the locBlock fields have changed, regardless of whether other
+     * form elements are dirty.
+     */
+    $scope.locBlockIsDirty = false;
+
+    /**
+     * This flag allows the code to distinguish between user- and
+     * server-initiated changes to the locBlock fields. Without this flag, the
+     * changes made to the locBlock fields when a location is fetched from the
+     * server would cause the watch function to mark the locBlock dirty.
+     */
+    $scope.locBlockSkipDirtyCheck = false;
+
     // The ts() and hs() functions help load strings for this module.
     var ts = $scope.ts = CRM.ts('org.civicrm.volunteer');
     var hs = $scope.hs = crmUiHelp({file: 'CRM/Volunteer/Form/Volunteer'}); // See: templates/CRM/volunteer/Project.hlp
@@ -166,6 +184,14 @@
     // VOL-223: Used to determine visibility of relationship block
     $scope.showRelationshipBlock = _.reduce($scope.showRelationshipType, function(a, b) {return (a || b); });
 
+    /**
+     * Populates locBlock fields based on user selection of location.
+     *
+     * Makes an API request.
+     *
+     * @see $scope.locBlockIsDirty
+     * @see $scope.locBlockSkipDirtyCheck
+     */
     $scope.refreshLocBlock = function() {
       if (!!$scope.project.loc_block_id) {
         crmApi("VolunteerProject", "getlocblockdata", {
@@ -186,6 +212,11 @@
     //Refresh as soon as we are up and running because we don't have this data yet.
     $scope.refreshLocBlock();
 
+    /**
+     * If the user selects the option to create a new locBlock (id = 0), set
+     * some defaults and display the necessary fields. Otherwise, fetch the
+     * location data so we can display it for editing.
+     */
     $scope.$watch('project.loc_block_id', function (newValue) {
       if (newValue == 0) {
         $scope.locBlock = {
@@ -203,7 +234,10 @@
       }
     });
 
-    //Watch the LocBlock and mark it if it as dirty if we modify it.
+    /**
+     * @see $scope.locBlockIsDirty
+     * @see $scope.locBlockSkipDirtyCheck
+     */
     $scope.$watch('locBlock', function(newValue, oldValue) {
       if ($scope.locBlockSkipDirtyCheck) {
         $scope.locBlockSkipDirtyCheck = false;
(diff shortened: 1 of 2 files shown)


TobiasLounsbury:
I refactored the whole dirty marker thing and turned it into a proper recursive watcher with flags for when we change out the whole block with new data so we only mark it dirty when we actually change a field.

ang/volunteer/Project.js
```diff
@@ -183,9 +185,15 @@
...
-      $scope.locBlockIsDirty = true;
-    };
+    //Watch the LocBlock and mark it if it as dirty if we modify it.
+    $scope.$watch('locBlock', function(newValue, oldValue) {
+      if ($scope.locBlockUpdating) {
+        $scope.locBlockUpdating = false;
+      } else {
+        $scope.locBlockIsDirty = true;
+      }
+      console.log("Dirty: ", $scope.locBlockIsDirty);
```
GinkgoFJG: Please remove debugging code.

ang/volunteer/Project.js
```diff
@@ -156,7 +156,9 @@
           "id": $scope.project.loc_block_id
         }).then(function(result) {
           if(!result.is_error) {
+            $scope.locBlockUpdating = true;
```
GinkgoFJG: I don't like the name of this flag. How about one of the following: locBlockFetch, locBlockSkipDirtyCheck?

GinkgoFJG:
Nice changes! This is much cleaner. We've drunk the Angular Kool-Aid!

Just a few changes and then I'll merge it. Thanks!

ginkgomzd:
Can we all go over this together?
    
    I have some questions and potential concerns if my understanding
    bears out.
    I think there might be some brute force happening here, for
    instance, allowing the listener to be called unconditionally.
    
    Double-bang??
https://github.com/civicrm/org.civicrm.volunteer/blob/99ad9eead53d61592d78eb03ee9412db61d7b213/ang/volunteer/Project.js#L152
    ... oh, I think I see... if is something that does not resolve to
    false? ...?
    
    Perhaps this should be broken out into multiple $watch listeners?
https://github.com/civicrm/org.civicrm.volunteer/blob/99ad9eead53d61592d78eb03ee9412db61d7b213/ang/volunteer/Project.js#L171
    ... $watch('project.loc_bloc_id == 0', function() {
    $scope.locBlock = {
              address: {
                country_id: _.findWhere(countries, {is_default: "1"}).id
              }
            };
    
            $("#crm-vol-location-block
    .crm-accordion-body").slideDown({complete: function() {
              $("#crm-vol-location-block
    .crm-accordion-wrapper").removeClass("collapsed");
    }});
    });
    ...
    why do we need to refreshLocBlock() every time $digest is called?
    
    It seems like we may want to take advantage of providing the
    ObjectEquality as the third parameter to $watch to do a check to
    locBlockUpdating.
    
    Another thing to consider is the variant which appears to be
    designed for our case...a complex object.
    $watchGroup()
    https://docs.angularjs.org/api/ng/type/$rootScope.Scope#!
    ... but maybe not, are users only selecting locations from a select
    list? i.e. not editing components of a locBlock?
    
    
    
    On 06/17/2016 07:18 PM, Frank J. Gómez
      wrote:
    
    
      @TobiasLounsbury,
        I made this change so that the code that was previously
        triggered by locBlockChanged (namely setting of the default
        country), could be triggered were the location block ID to be
        changed before the form is rendered (i.e., by crmHook).
      
      You can view, comment on, or merge this pull request online
        at:
        https://github.com/civicrm/org.civicrm.volunteer/pull/392
      Commit Summary
      
        Refactored so that handler code for changes to the loc block
          selection is
      
      File Changes
      
         M ang/volunteer/Project.html
          (2) 
         M ang/volunteer/Project.js
          (7) 
      
      Patch Links:
      
        https://github.com/civicrm/org.civicrm.volunteer/pull/392.patch
        https://github.com/civicrm/org.civicrm.volunteer/pull/392.diff
      
      —
        You are receiving this because you are subscribed to this
        thread.
        Reply to this email directly, view
          it on GitHub, or mute
          the thread.

TobiasLounsbury:
@ginkgomzd  Where are we refreshing the locBlock on every digest? The way it should work is that whenever the dropdown with locations changes, the refresh function pings the server and fetches the data associated with that location and makes it available to edit in the (accordion) form block. The dirty flag is separate because we need to know if something about a given locBlock has changed. If you change the location of a project, we don't need to update that location block, unless you also updated the name of the street for example. The id zero check is because we must create the locBlock BEFORE the project is saved. 

The double bang is a javascript convention that is the same as casting the variable to a Boolean.

ginkgomzd:
@TobiasLounsbury  I was misunderstanding the ObjectEquality param and your use of it.
https://github.com/civicrm/org.civicrm.volunteer/blob/99ad9eead53d61592d78eb03ee9412db61d7b213/ang/volunteer/Project.js#L189
I see now that the listener is not going to always be invoked.

okay, don't know when we'll get to meet. I created a doc to summarize what I'm seeing and help us formalize this pattern.
https://docs.google.com/document/d/1aTK7te0ZKVn4f9g4y_EV_8qvbv5VYY5cgTJlpWxFT5E/edit#

...TBC

GinkgoFJG:
This looks good to me. @ginkgomzd, do you want to discuss this more before it's merged?

TobiasLounsbury:
@ginkgomzd To be fair, Angular keeps track of form status for us. It will mark fields as dirty at the form level. However that pattern didn't work because we are injecting the locBlock data after the fact, and there wasn't a way I could find to mark a field as clean after you had just changed it, without marking the whole form as clean. Thus I created this custom locBlockIsDirty flag that only deals with the locBlock sub-form. the project screen is really two forms, one nested in the other. The project is one form, and then the location Block is a separate form, that has to be saved and updated apart from the project data. That is why we have a custom dirty handler mechanism.

ginkgomzd:
@GinkgoFJG I leave it to you to decide when the PR is ready to merge. I do think there is more to discuss here. Toby's last comment adds to my understanding and shows that a guiding doc for long-term concerns is not yet complete.
I'm not yet clear on how/if future maintenance might stumble on this. If this is outside of the standard angular paradigm, some more "why" comments might be needed. I still wonder if the code could be more concise by changing watch expressions... I want to question also the setting of locBlockDirty in saveProject but again maybe a comment to explain which case that is covering compared to the other places.

Shall we make it a show-and-tell session this Friday? Pair-documentation maybe.

PR #400, opened 2016-09-30
VOL-248: Display custom project relationship fields on the project create/edit screen.

GinkgoFJG:
* [VOL-248: Regression: Custom project relationship fields are not displayed on project create\/edit screen](https://issues.civicrm.org/jira/browse/VOL-248)

### DIFF
--- a/ang/volunteer/Project.js
+++ b/ang/volunteer/Project.js
@@ -112,6 +112,11 @@
       });
     }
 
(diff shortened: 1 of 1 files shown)


GinkgoFJG:
This code is getting a little hairy and the logic for how to display these fields is getting complex enough that it may be time to encapsulate it... but here's a quick fix.

PR #416, opened 2016-11-12
VOL-202: Bugfixes for newly implemented updates.

GinkgoFJG:
With these commits, the following manual user tests pass:

* create via API as admin with just a title, expect all defaults applied
* update via GUI as admin with new project contacts, expect only project contacts to change
* update via GUI as admin with new profiles, expect only profiles to change
* update via API as admin with new project contacts, expect only project contacts to change
* update via API as admin with new profiles, expect only profiles to change
* create via GUI as admin, expect defaults not to override supplied values
* create via API as coord with just a title, expect all defaults applied
* create via GUI as coord, expect project contact and profiles supplied
* update via GUI as coord, expect project contact and profiles unchanged
* update via API as coord, expect project contact and profile params to be ignored
* create via API as coord, expect project contact and profile params to be ignored, defaults instead

... where user admin has "edit volunteer project relationships" and "edit volunteer registration profiles" and user coord does not.

---

 * [VOL-202: Add Project Relationship Defaults to Defaults UI](https://issues.civicrm.org/jira/browse/VOL-202)

### DIFF
--- a/CRM/Volunteer/BAO/Project.php
+++ b/CRM/Volunteer/BAO/Project.php
@@ -187,10 +187,10 @@ class CRM_Volunteer_BAO_Project extends CRM_Volunteer_DAO_Project {
     if (!array_key_exists('loc_block_id', $params)) {
       $params['loc_block_id'] = $defaults['loc_block_id'];
     }
-    if (!array_key_exists('profiles', $params)) {
+    if (!array_key_exists('profiles', $params) || !CRM_Volunteer_Permission::check('edit volunteer registration profiles')) {
       $params['profiles'] = $defaults['profiles'];
     }
-    if (!array_key_exists('project_contacts', $params)) {
+    if (!array_key_exists('project_contacts', $params) || !CRM_Volunteer_Permission::check('edit volunteer project relationships')) {
       $params['project_contacts'] = $defaults['relationships'];
     }
 
--- a/CRM/Volunteer/BAO/Project.php
+++ b/CRM/Volunteer/BAO/Project.php
@@ -299,6 +299,11 @@ class CRM_Volunteer_BAO_Project extends CRM_Volunteer_DAO_Project {
         if (is_array($profile['module_data'])) {
           $profile['module_data'] = json_encode($profile['module_data']);
         }
+
+        // Since we delete then recreate the ufjoins, drop the ID from the params
+        // or else CiviCRM tries to update a nonexistent record and fails silently
+        unset($profile['id']);
+
         civicrm_api3('UFJoin', 'create', $profile);
       }
     }
--- a/CRM/Volunteer/BAO/Project.php
+++ b/CRM/Volunteer/BAO/Project.php
@@ -291,32 +291,36 @@ class CRM_Volunteer_BAO_Project extends CRM_Volunteer_DAO_Project {
     }
 
     if ($updateVPC || CRM_Core_Action::ADD) {
-      foreach ($params['project_contacts'] as $relationshipType => $contactIds) {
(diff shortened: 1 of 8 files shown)


CRM/Volunteer/BAO/Project.php
```diff
@@ -187,10 +187,10 @@ public static function supplyDefaults(array $params) {
     if (!array_key_exists('loc_block_id', $params)) {
       $params['loc_block_id'] = $defaults['loc_block_id'];
     }
-    if (!array_key_exists('profiles', $params)) {
+    if (!array_key_exists('profiles', $params) || !CRM_Volunteer_Permission::check('edit volunteer registration profiles')) {
```
ginkgomzd: I have to raise a flag for mixing concerns. Is it a convenience to do it here? That is, the default is only used if the permission-check fails AND the parameter is not already set? One advantage of having a "single responsibility principle" function for permissions could be optimising the permissions look-up. Permissions checks could involve group-look-ups, no? I see it is a static function. Is there a singleton that caches the look-up?

CRM/Volunteer/BAO/Project.php
```diff
@@ -250,10 +250,22 @@ public static function create(array $params) {
 
     $project->save();
 
+    // Prevent updates for unpermissioned users.
+    $hasEditVPC = CRM_Volunteer_Permission::check('edit volunteer project relationships');
```
ginkgomzd: $canEditVPContacts

CRM/Volunteer/BAO/Project.php
```diff
@@ -250,10 +250,22 @@ public static function create(array $params) {
...
+    $hasEditProfiles = CRM_Volunteer_Permission::check('edit volunteer registration profiles');
```
ginkgomzd: $canEditVPProfiles

ginkgomzd: Doesn't this make the permissions check in supplyDefaults unnecessary?

GinkgoFJG: I don't think so. I didn't like checking the permissions more than once, which is in part what you complained about above, but the checks serve different purposes.

In `supplyDefaults` they decide whether to honor the user submission or force the defaults. In `create` they determine whether to write the ancillary entities at all.

CRM/Volunteer/BAO/Project.php
```diff
@@ -269,13 +292,18 @@ public static function create(array $params) {
...
+          // For some reason, the ID param implicit in chained API calls is not
+          // getting passed, so we specify it manually. Since the API is
+          // deprecated, this short-term solution is good enough.
+          'id' => '$value.id',
+        ),
       ));
     }
 
-    if ($updateVPC || CRM_Core_Action::ADD) {
+    if ($updateVPC || $op === CRM_Core_Action::ADD) {
```
ginkgomzd: I think this is evidence of how code suffers when you mix concerns. This test should be "was project_contacts passed?" but you can't do that because supply_defaults pollutes the params.

I think this function is bending over backwards to implement implicit deletes for project-relationships and uf_joins. Moving the logic for ufjoins and project-relationships to separate functions would make this read much more easily. It might then be okay to not have distinct default functions which might be more transparent. A prototypical create function could be: check permissions; validate; default; ensure child-objects; persist.

I know I'm belabouring this function, but I feel like bad idioms in CiviCRM are infecting the code and I see how this could get ugly down the road as Project entities get more complex, or we add finer-grained permissions, etc. Dependant lines of code are getting separated and will get more separated as the code grows. Ideally, I prefer to see a logic-branch just once. Branching on the same test is a red-flag for maintainability and a good indicator of mixing concerns.

PR #419, opened 2016-11-22
Fixed fatal error for PHP 5.6

Edzelopez:
Adding volunteer hours results in following error message and WSOD. 
Cannot access self:: when no class scope is active


GinkgoFJG:
I am unable to reproduce this, @Edzelopez. Could you please create an issue in JIRA with steps for reproducing the problem? Thanks.

PR #420, opened 2016-11-25
VOL-247 VOL-251 - Fix New or No location in project edit.

agileware:
* [VOL-247: Creating a project with no location gives a fatal error](https://issues.civicrm.org/jira/browse/VOL-247)
 * [VOL-251: Unable to add location to project](https://issues.civicrm.org/jira/browse/VOL-251)

### DIFF
--- a/ang/volunteer/Project.js
+++ b/ang/volunteer/Project.js
@@ -360,13 +360,11 @@
      */
     saveProject = function() {
       if ($scope.validateProject()) {
-        if($scope.project.loc_block_id == "0" || $scope.project.loc_block_id == "") {
+        if($scope.project.loc_block_id == "0") {
           delete $scope.project.loc_block_id;
         }
-	 
-        
+
         return crmApi('VolunteerProject', 'create', $scope.project).then(function(result) {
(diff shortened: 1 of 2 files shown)


GinkgoFJG:
Hi, Justin. Thanks for submitting this patch. I've been working on this same issue, and I have a different patch which I'm just about ready to submit as a PR which I like a little better.

One of the things I prefer about the approach in the forthcoming PR is that is it's a bit less conservative than yours in cleaning up cruft/hasty architecture. You've managed to get rid of the "placeholder" locblock, which is great, but I think I've also managed to get rid of the need to save projects twice (once before the locblock, and once after).

GinkgoFJG:
For what it's worth, I think these issues were introduced by a change in CiviCRM core sometime during the 4.7 series. Mitigating the blame, however, is the fact that Volunteer was using an internal BAO function where a public API would have worked.

Have a look at my alternative PR #421, please, and let me know what you think. I haven't tested it against 4.6, though I don't anticipate any problems there.

agileware:
Superseded by #421

PR #421, opened 2016-11-29
Fix for location issues VOL-247 and VOL-251

GinkgoFJG:
* [VOL-247: Creating a project with no location gives a fatal error](https://issues.civicrm.org/jira/browse/VOL-247)
 * [VOL-251: Unable to add location to project](https://issues.civicrm.org/jira/browse/VOL-251)


agileware:
Late to the party, but works for me

PR #427, opened 2016-12-06
VOL-273: Fix for volunteer projects not saving due to reversed boolean.

GinkgoFJG:


GinkgoFJG:
FYI, @ginkgomzd, I merged this one myself because I wanted to get 2.2.1 out quickly. The effects of this bug are rather substantial.

PR #428, opened 2016-12-15
VOL-271 get list params function for volunteer needs

alifrumin:
* [VOL-271: getlist api functionality for volunteer needs](https://issues.civicrm.org/jira/browse/VOL-271)

### DIFF
--- a/api/v3/VolunteerNeed.php
+++ b/api/v3/VolunteerNeed.php
@@ -208,5 +208,4 @@ function _civicrm_api3_volunteer_need_getlist_params(&$request) {
   );
   $request['params']['return'] = array_unique(array_merge($fieldsToReturn, $request['extra']));
   $request['params']['options']['sort'] = 'start_time DESC';
-  );
 }
--- a/api/v3/VolunteerNeed.php
+++ b/api/v3/VolunteerNeed.php
@@ -208,7 +208,5 @@ function _civicrm_api3_volunteer_need_getlist_params(&$request) {
   );
   $request['params']['return'] = array_unique(array_merge($fieldsToReturn, $request['extra']));
   $request['params']['options']['sort'] = 'start_time DESC';
-  $request['params'] += array(
-    'is_active' => 1,
   );
 }
--- a/api/v3/VolunteerNeed.php
+++ b/api/v3/VolunteerNeed.php
@@ -186,3 +186,29 @@ function civicrm_api3_volunteer_need_delete($params) {
   return _civicrm_api3_basic_delete('CRM_Volunteer_BAO_Need', $params);
 }
 
+/**
+ * Get parameters for Volunteer Need list.
+ *
+ * @see _civicrm_api3_generic_getlist_params
+ *
+ * @param array $request
+ *   API request.
+ */
+function _civicrm_api3_volunteer_need_getlist_params(&$request) {
+  $fieldsToReturn = array(
+    'role_label',
+    'role_id',
+    'display_time',
+    'project_id',
+    'start_time',
+    'duration',
+    'is_flexible',
+    'visibility_id',
+    'quantity',
+  );
+  $request['params']['return'] = array_unique(array_merge($fieldsToReturn, $request['extra']));
+  $request['params']['options']['sort'] = 'start_time DESC';
+  $request['params'] += array(
+    'is_active' => 1,
+  );
+}


GinkgoFJG:
Hi, Alice. Thanks for submitting this.

I have to say I find the magic functions around the getList API a little confusing. I just today tangoed with the `_civicrm_api3_{$entity}_getlist_defaults` function and [documented my findings](https://wiki.civicrm.org/confluence/display/CRMDOC/EntityRef+Fields#EntityRefFields-_civicrm_api3_{$entity}_getlist_defaults). From my limited understanding, I think you may be mixing concerns with the `is_active` part of your patch; it looks to me like you're setting a default there.

In any case, could you explain a bit more how this works and what the change enables you to do that you couldn't do before?

Aside: @colemanw, I documented some of my findings today re `_civicrm_api3_{$entity}_getlist_defaults` (link above), but it's unclear to me what the three different customization points are for. Could you please review my doc (make sure it isn't more harmful than helpful!), and if it's not too much trouble fill the blanks for the `_getlist_params` and `_getlist_output` functions? It's just not clear to me which one should be used when for what. Thanks!

alifrumin:
Hey Frank,

I found the magic functions around the getList API a little confusing too. I based the is_active part of the patch off of the [function _civicrm_api3_event_getlist_params](https://github.com/civicrm/civicrm-core/blob/master/api/v3/Event.php#L232), but somewhere along the way the if surrounding it got lost. I  will take out that line. 

I was working to set up a select 2 for Vounteer Needs using .crmEntityRef like so:

```js

      $('#needSelectId').crmEntityRef({
        entity: 'VolunteerNeed',
        api: {
          params: { project_id: $volProjectId },
          description_field: ['display_time', 'role_description'],
          label_field: 'role_label',
        },
      });

```

I found that the [generic getlist functions](https://github.com/civicrm/civicrm-core/blob/master/api/v3/Generic/Getlist.php) do not work for Volunteer Needs. This PR approaches creating a working getlist api for Volunteer Needs by explicitly listing the fields to return because some of the fields in the  [function civicrm_api3_volunteer_need_get](https://github.com/civicrm/org.civicrm.volunteer/blob/master/api/v3/VolunteerNeed.php#L75) (namely display_time, role_label, and role_description) are not in the Volunteer Need DAO and specifying the sort order because the getlist api [defaults to sorting by the label field](https://github.com/civicrm/civicrm-core/blob/master/api/v3/Generic/Getlist.php#L121) which is not guaranteed to be in the Volunteer Need DAO either (as you can see in the example above I am using the role label as the label_field). These changes make it so one can use .crmEntityRef to generate an infinitely scrolling select2 for Volunteer Needs.

Some other approaches we considered to put together a working getList api for Volunteer Need were to add the fields (display_time, role_label, and role_description) to the DAO or to create a function that generates a table that holds all the fields and use that instead. 

I submitted a [different PR to make the project_id a pseudo constant](https://github.com/civicrm/org.civicrm.volunteer/pull/429) so one could use crmEntityRef to generate a select2 that looked something like "Project Title - Role Label" as well. Hope this clears things up some, let me know if you have more questions.

api/v3/VolunteerNeed.php
```diff
@@ -186,3 +186,26 @@ function civicrm_api3_volunteer_need_delete($params) {
...
+  $fieldsToReturn = array(
+    'role_label',
+    'role_id',
+    'display_time',
+    'project_id',
+    'start_time',
+    'duration',
+    'is_flexible',
+    'visibility_id',
+    'quantity',
```
colemanw: What are all these return params for?

alifrumin: I just had it return all the fields one can access from the volunteer need get api, we could definitely cut out a bunch of them

colemanw:
Hmm, typically there is only about one "right" set of params for an entity ref, so if the defaults don't work, change 'em. In your code example above, you had to manually set the "description_field" and "label_field" every time you create an EntityRef. That should be unnecessary. Instead set those as the defaults by declaring a `_civicrm_api3_volunteer_need_getlist_defaults` function. You can also specify sort order in that same function, so I think this PR is probably unnecessary.

alifrumin:
I see what you are saying, I will create a pr for a _civicrm_api3_volunteer_need_getlist_defaults function instead

colemanw:
Note that function does not need to accept any args, and it returns an array.

alifrumin:
Thanks @colemanw https://github.com/civicrm/org.civicrm.volunteer/pull/431

PR #429, opened 2016-12-15
VOL-275 making the volunteer project a pseudoconstant so one can more easily access it from vol needs

alifrumin:
* [VOL-275: make the project title easily accessible in the get and getlist api for volunteer needs ](https://issues.civicrm.org/jira/browse/VOL-275)

### DIFF
--- a/CRM/Volunteer/DAO/Need.php
+++ b/CRM/Volunteer/DAO/Need.php
@@ -185,6 +185,17 @@ class CRM_Volunteer_DAO_Need extends CRM_Core_DAO
           'description' => 'FK to civicrm_volunteer_project table which contains entity_table + entity for each volunteer project (initially civicrm_event + eventID).',
           'required' => false,
           'FKClassName' => 'CRM_Volunteer_DAO_Project',
+          'title' => ts('Project') ,
+          'import' => TRUE,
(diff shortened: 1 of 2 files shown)


ginkgomzd:
@colemanw  do you have some input here? I am reading into the intent that the desire is to achieve a performance gain through the caching of pseudoconstants, otherwise why not do a regular project title retrieval... 
I know some of our volunteer instances have a relative high-usage of projects and I worry that this performance gain would be negated at times. I'm always wary of premature optimization. I think the definition on the wiki of Pseudoconstants is pretty clear and this case is not so clear because whether or not volunteer Projects meet the criteria of changing infrequently is up for discussion. I also wonder if there is a risk in having very large option lists.

@alifrumin I see you are clearly in the middle of something... would you be interested in discussing the use-case with us and discussing architecture? We recently created a vision for some significant changes driven by what I think is a similar problem... that Needs have little context on their own.

alifrumin:
@GinkgoFJG I was working on this https://github.com/aghstrategies/com.aghstrategies.logvolunteertime but don't need this pr accepted or the one regarding the volunteer need getlist to make it work, they were things I stumbled upon as I was working on t that I thought might be helpful to others. If they don't fit with your guys vision don't sweat it.

PR #430, opened 2016-12-22
VOL-276: Path to deprecating api.VolunteerUtil.getbeneficiaries.

GinkgoFJG:
* [VOL-276: Deprecate api.VolunteerUtil.getbeneficiaries](https://issues.civicrm.org/jira/browse/VOL-276)

### DIFF
--- a/ang/volunteer/Projects.html
+++ b/ang/volunteer/Projects.html
@@ -1,4 +1,5 @@
 <div class="crm-container">
+  <!-- TODO for VOL-267: Remove beneficiaries debugging block. -->
   <div crm-ui-debug="beneficiaries"></div>
   <div crm-ui-debug="campaigns"></div>
   <div crm-ui-debug="projects"></div>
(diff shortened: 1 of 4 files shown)


GinkgoFJG:
The [`formatBeneficiaries()`](https://github.com/civicrm/org.civicrm.volunteer/pull/430/files#diff-d0685484f6e18c925e8aa3c71af27442L133) method is a little problematic because I was intending to factor out the [`beneficiaries`](https://github.com/civicrm/org.civicrm.volunteer/pull/430/files#diff-d0685484f6e18c925e8aa3c71af27442L11) promise altogether.

The primary function of the `getList` APIs is to support CiviCRM's `entityRef` widget. Because the list is pulled on demand when the user interacts with the widget, there shouldn't be any need to pre-load the data. I didn't realize right away that `formatBeneficiaries()` uses the resolved promise to translate the IDs returned in the `projectData` promise for use in the beneficiaries column on the Manage Projects screen.

I *could* replace the deprecated API call in the `beneficiaries` promise with the new one, but I think it would be preferable to use the API's newish join capabilities to fetch the display name in the `projectData` promise and eliminate the extraneous `beneficiaries` promise once and for all.

PS: Looks like I missed a [TODO](https://github.com/GinkgoFJG/org.civicrm.volunteer/blob/1d1e68189ce5a10f987f22ce612ce87cda82685e/ang/volunteer/Projects.html#L88).

ginkgomzd:
I don't have strong feelings about removing the beneficiaries promise from the angular page. It's less of an api. Could serve as useful guidance to developers building something custom. Just playing devil's advocate. If you want to move forward with removing the api call, I think you can.

PR #431, opened 2016-12-29
VOL-271 Set defaults for Volunteer Need get list

alifrumin:
* [VOL-271: getlist api functionality for volunteer needs](https://issues.civicrm.org/jira/browse/VOL-271)

### DIFF
--- a/api/v3/VolunteerNeed.php
+++ b/api/v3/VolunteerNeed.php
@@ -186,3 +186,24 @@ function civicrm_api3_volunteer_need_delete($params) {
   return _civicrm_api3_basic_delete('CRM_Volunteer_BAO_Need', $params);
 }
 
+/**
+ * Set the default getList behavior for Volunteer Need to have the label be the role
+ * and the description be the formatted date and time
+ *
+ * @param array $request
+ *   The parameters passed to the sub-API call (i.e., the parameters to the get
+ *   call underlying the getList call). These are passed to getList in
+ *   $params['params'].
+ * @return array
+ *   Despite the fact that $request represents a subset of the parameters passed
+ *   to getList, the return of this function is merged with the getList params
+ *   in their entirety.
+ */
+function _civicrm_api3_volunteer_need_getlist_defaults(&$request) {
+  return array(
+    'extra' => array('is_flexible', 'role_id', 'role_label', 'start_time', 'display_time'),
+    'params' => array('options' => array('sort' => 'start_time DESC')),
+    'label_field' => 'role_label',
+    'description_field' => array('display_time'),
+  );
+}


api/v3/VolunteerNeed.php
```diff
@@ -186,3 +186,24 @@ function civicrm_api3_volunteer_need_delete($params) {
...
+ *   call underlying the getList call). These are passed to getList in
+ *   $params['params'].
+ * @return array
+ *   Despite the fact that $request represents a subset of the parameters passed
+ *   to getList, the return of this function is merged with the getList params
+ *   in their entirety.
+ */
+function _civicrm_api3_volunteer_need_getlist_defaults(&$request) {
+  return array(
+    'extra' => array('is_flexible', 'role_id', 'role_label', 'start_time', 'display_time'),
```
colemanw: Setting a default "extra" is probably not needed. This is a rarely used (but handy) feature that allows you to return extra hidden data for use in javascript. Typically you'd set this param on a per-instance basis in your entityRef field, rather than declaring a default.

alifrumin:
It breaks if you don't include the extra fields.

api/v3/VolunteerNeed.php
alifrumin: It breaks if you don't include the extra fields. Specifically 'role_label' and 'role_description' display as null unless they and 'role_id' are passed as extra and 'display_time' always displays as 'Any' unless you include start_time and 'is_flexible' makes it so if there is no 'role_label' it displays 'Any' instead of null.

colemanw: Hmm, ok that's odd. The function `_civicrm_api3_generic_getlist_params` ought to be automatically including those.

alifrumin: That's why I started with the civicrm_api3_volunteer_need_getlist_params, maybe I didn't explain it well enough. A bunch of fields are assembled in the volunteer need get [here](https://github.com/civicrm/org.civicrm.volunteer/blob/master/api/v3/VolunteerNeed.php#L75), so in order to access them from volunteer need getlist api I found one could add them as "extra"  fields or add them to the array of fields to return in the_civicrm_api3_generic_getlist_params function

colemanw: Makes sense. A slightly less hacky way I suppose would be to modify the `get` function to include the fields required for the calculated fields, e.g. if you specify `display_time` in the $return params, it should auto-include `start_time, duration, end_time`. But don't forget that $return can be a comma-separated string as well as an array, so kind of a pain to work with.

alifrumin: Hadn't thought of that thanks for the tips

PR #432, opened 2017-01-10
VOL-282: Removed two avenues which allowed over-registration of volunteer opps

GinkgoFJG:
* [VOL-282: Volunteers can register for filled opportunities](https://issues.civicrm.org/jira/browse/VOL-282)

### DIFF
--- a/CRM/Volunteer/Form/VolunteerSignUp.php
+++ b/CRM/Volunteer/Form/VolunteerSignUp.php
@@ -69,6 +69,13 @@ class CRM_Volunteer_Form_VolunteerSignUp extends CRM_Core_Form {
   protected $_needs = array();
 
   /**
+   * Error messages to display, related to the need IDs passed to the form via URL.
+   *
+   * @var array
+   */
+  protected $preProcessErrors = array();
+
+  /**
    * The profile IDs associated with this form and marked
    * for use with the primary contact.
    *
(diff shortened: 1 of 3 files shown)


CRM/Volunteer/Form/VolunteerSignUp.php
```diff
@@ -706,4 +768,20 @@ protected function setDestination() {
     $this->_destination = CRM_Utils_System::url($path, $query, FALSE, $fragment);
   }
 
+  /**
+   * If the form preprocessing uncovers problems with the passed needs, display
```
ginkgomzd: Don't comment with business logic.
"Sub-routine of buildQuickForm to display validation messages to the user."

CRM/Volunteer/Form/VolunteerSignUp.php
```diff
@@ -317,6 +369,16 @@ function buildQuickForm() {
     if ($allowAdditionalVolunteers) {
       //Give the volunteer a box to select how many friends they are bringing
       $this->add("text", "additionalVolunteerQuantity", ts("Number of Additional Volunteers", array('domain' => 'org.civicrm.volunteer')), array("size" => 3));
+
+      // VOL-282: Cap how many additional volunteers can be added based on the opp with the fewest openings
+      $quantitiesAvailable = array_column($this->_needs, 'quantity_available');
```
ginkgomzd: ... didn't know there was such a thing as array_column()

GinkgoFJG: It's my new hammer.

GinkgoFJG:
You're wondering about the UX and find it odd that "This project is not active" and "Sorry, this project is full" are given the same severity. I'm not sure what you mean. I think the "An unrecoverable error has occurred" page title is a bit heavy-handed and would appreciate alternative suggestions. However, I think these two scenarios have the same outcome: we want to prevent sign-ups and suggest alternate actions to the user. What am I not getting?

ginkgomzd:
Maybe I'm assuming that one is much more likely to occur than the other. How likely is it that a user is trying to sign up for an inactive project? An error hasn't occurred, has it? "Please go back and make a different selection".

GinkgoFJG:
You're not convincing me anything should change. I think the scenarios are equally likely:

1. An organization creates a volunteer opportunity and emails a link directly to the signup form, bypassing the shopping cart.
2. In the three days it takes for Joe Smith to get through his email backlog, find the email from the org asking him to register, and click the link:
    * the project has been disabled, or
    * all the openings for the opportunity have been filled by people with better e-hygiene than Joe.

I'll update the code comments and user-facing language then merge, unless you still have concerns.

ginkgomzd:
okay, I see why you are handling them together. If you agree the UX language should change, then we are in agreement.

PR #437, opened 2017-02-07
VOL-287: Replace opportunities settings stub with API call

agileware:
https://issues.civicrm.org/jira/browse/VOL-287

### DIFF
--- a/CRM/Volunteer/Angular/Manager.php
+++ b/CRM/Volunteer/Angular/Manager.php
@@ -95,6 +95,14 @@ class VolunteerManager extends Manager {
       $this->modules = $this->resolvePatterns($angularModules);
     }
 
+    $cvsetting = civicrm_api('Setting', 'getsingle', array(
+                   'version' => 3,
+                   'group' => 'org.civicrm.volunteer',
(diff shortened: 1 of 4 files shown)


ginkgomzd:
Looks good to me

agileware:
Should've tested better as non admin - of course you can't query the settings API from javascript unless you have "administer civicrm".

GinkgoFJG:
@agileware, thanks for fleshing out this feature by adding actual settings.

We already have a hack in place for addressing the problem of the permissions needed for the various calls involved in building an Angular page. We aren't proud of it, but one hack pattern is clearly preferable to a new pattern each time there's a new problem to solve.

The `supporting_data` resolve property makes a call to `api.VolunteerUtil.supporting_data`. This is where we've stashed other junk needed to render a page. For the controller in question, you'll see the roles are fetched this way. The pattern is consistent across other Angular-based CiviVolunteer pages.

Could you please re-work your logic to fit within the established pattern?

Thanks!

jusfreeman:
@GinkgoFJG feel free to close this PR or pick it up and run with it. We are no longer supporting any organisations using CiviVolunteer.

PR #439, opened 2017-02-08
VOL-187: Fixes bug when searching by more than one beneficiary.

GinkgoFJG:
* [VOL-187: Error: "Beneficiary is not a valid integer" when searching by more than one beneficiary](https://issues.civicrm.org/jira/browse/VOL-187)

### DIFF
--- a/ang/volunteer.js
+++ b/ang/volunteer.js
@@ -95,6 +95,11 @@
        * @return string
        */
       var buildQueryString = function () {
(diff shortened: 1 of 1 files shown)


GinkgoFJG:
@ginkgomzd: FYI, I branched the code base. New development promises compatibility with 4.7 only. Support for 4.6 is bugfix-only.

PR #440, opened 2017-02-09
VOL-288: Fixed install bug: CiviCRM_API3_Exception: "N is not a valid…

GinkgoFJG:
… option for field option_group_id"

### DIFF
--- a/CRM/Volunteer/Upgrader.php
+++ b/CRM/Volunteer/Upgrader.php
@@ -131,6 +131,9 @@ class CRM_Volunteer_Upgrader extends CRM_Volunteer_Upgrader_Base {
         'is_active' => 1,
       ));
       $optionGroupId = $optionGroup['id'];
+
+      // VOL-288: Prevent caching-related CiviCRM_API3_Exception: "N is not a valid option for field option_group_id"
+      CRM_Core_Invoke::rebuildMenuAndCaches();
     } catch (Exception $e) {
       // if an exception is thrown, most likely the option group already exists,
       // in which case we'll just use that one


GinkgoFJG:
Hey, @totten. This works, but it feels wrong. Is there a better way to do this? TLDR from the ticket:

If you're programmatically creating an optionGroup and adding an optionValue to it, you have to clear caches between steps 1 and 2 or you'll get the error above.

totten:
Yeah, it feels excessive to do a full rebuild. Assuming the problem is as-described, it would be more proportionate to say, e.g., "whenever you add/edit/delete an OptionGroup or OptionValue, flush the related caches".

The challenge would be figuring out where those caches are. I don't know this specific bug, but on general grounds these places might be implicated:
 * `CRM_Core_OptionGroup` (has some unintuitive double-caching)
 * `CRM_Core_PseudoConstant` (has caching and has been an issue in the past; would expect it to have been beaten into shape by now, but you never know)
 * `CRM_Core_BAO_OptionGroup` (sounds plausible but don't see any caching logic)
 * `CRM_Core_BAO_OptionValue` (eg has some caching in `getOptionValuesArray`)

You might try doing a few smaller/more targeted flushes (e.g. `CRM_Core_Config::singleton()->cleanupCaches()`)

ginkgomzd:
@GinkgoFJG @totten I propose we merge this. Excessive cache clearing on enabling the extension seems like a minor sin and no one is jumping for the opportunity to make it more selective.
Thoughts?

totten:
@ginkgomzd - Yes, IMHO, it's better for the extension to fire `rebuildMenuAndCaches()` than to let an installer bug sit around. I wouldn't recommend it as a default/pro-forma practice for all extensions, but it's a valid work-around if it's needed.

Aside: [this issue in mosaicomsgtpl](https://github.com/civicrm/org.civicrm.mosaicomsgtpl/issues/3) is another example of firing `rebuildMenuAndCaches()`.  It's not necessary to read/understand for resolving this one, but it could give some additional perspective.

PR #441, opened 2017-02-09
VOL-288: Fixed install bug: CiviCRM_API3_Exception: "N is not a valid…

universalhandle:
… option for field option_group_id"


GinkgoFJG:
Merge if #440 is merged.

PR #442, opened 2017-02-10
VOL-278: Enable Customizing the UF Page Title.

agileware:
Use page title for Angular pages that matches the pattern that crmUi matches.
Add page title replacement for (user-facing) opportunities search page.

---

 * [VOL-278: Enable Customizing the UF Page Title](https://issues.civicrm.org/jira/browse/VOL-278)

### DIFF
--- a/ang/volunteer/VolOppsCtrl.html
+++ b/ang/volunteer/VolOppsCtrl.html
@@ -1,7 +1,8 @@
 <!--
 Controller: VolOppsCtrl
 Required vars: volOppData(), searchParams
--->
+   -->
+<h1 crm-page-title>{{ts('Search for Volunteer Opportunities')}}</h1>
 
 <div class="crm-container">
   <div crm-ui-debug="volOppData()"></div>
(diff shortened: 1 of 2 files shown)


GinkgoFJG:
Looks like this needs some more work. Now every page in the volunteer Angular app gets the title "Search for Volunteer Opportunities." Also, why did you make the change to the XML file? It seems superfluous.

agileware:
@GinkgoFJG the change to the XML file is needed for crmUi's crmPageTitle directive to work - it looks for the known page title (which has been hardcoded in crmUi to start out as 'CiviCRM') and will *only* replace that.

In our testing this only sets the title to "Search for Volunteer Opportunities" for the /volunteer/opportunities route -- will need to re-test the event pages but other parts of the angular app should say "CiviCRM"

GinkgoFJG:
A read of the directive code suggests that's not quite accurate. It [starts off](https://github.com/civicrm/civicrm-core/blob/master/ang/crmUi.js#L5) with the assumption that the page title is "CiviCRM," but later that page title is set to -- well, whatever [you set it to](https://github.com/civicrm/civicrm-core/blob/master/ang/crmUi.js#L5).

I ran into the problem by loading up the search vol opps screen then loading up the manage project screen. I suspect the problem can be resolved by using the directive on each route (i.e., explicitly setting page titles everywhere).

jusfreeman:

PR #443, opened 2017-02-10
VOL-289: Log Volunteer Hours form should use custom Date format

agileware:
* [VOL-289: Log Volunteer Hours form should use custom Date format](https://issues.civicrm.org/jira/browse/VOL-289)


GinkgoFJG:
Hello, @agileware,

Thanks for this pull request! Your suggested fix results in the "frozen" fields and the editable fields using different date formats. My update to your pull request ensures both use the "activityDateTime" format, however the user has that configured.

Note: Since I've gone ahead and factored out the deprecated `$this->addDateTime()`, this pull request is good for 4.7 only and hence will not be merged into branch `4.6-2.2`.

PR #444, opened 2017-02-13
VOL-290: Replace mailto: link with activity pop-up in Volunteer roster

agileware:
* [VOL-290: Volunteer roster should send email from CiviCRM System](https://issues.civicrm.org/jira/browse/VOL-290)


GinkgoFJG:
@ginkgomzd, holding off on code review until we reach consensus on how this screen should work, per the thread on [VOL-230](https://issues.civicrm.org/jira/browse/VOL-230).

jusfreeman:

ginkgomzd:
This is a significant change in the original use-case for the volunteer roster. I don't think this is appropriate to be upstreamed.

PR #456, opened 2017-08-16
Update Projects.html

wildsight:
Adds a Volunteer Signup Page link to each project on the manage volunteer projects page, like the Event Signup links found on the Manage Events page. This was requested after a user spent a long time trying to figure this out.


GinkgoFJG:
Hi, @wildsight ,

Sorry for the delayed response, and thanks for this patch. What do you think about making the link open in a new tab, like Manage Events workflow you mentioned? Seems less jarring than going from an admin view to an end-user view.

And a second thought: there's another poorly documented URL parameter you might be interested in. For the link you created, you have the option to hide the search fields:

- to disable them: #/volunteer/opportunities?project=1&hideSearch=always
- to disable them with the option to restore them: #/volunteer/opportunities?project=1&hideSearch=1

Any opinion on the matter?

wildsight:
I'd definitely support new tab and strongly support hideSearch=1.

GinkgoFJG:
Superceded by #469.

PR #459, opened 2017-09-15
VOL-312: TFC-94 - Add more detail to volunteer activity when assigned

agileware-fj:
Change fields of activity set when creating from 'Assign Volunteers' page.

1. "With Contact" field is set to be the Volunteer Project, "Manager" instead of "Beneficiary".
2. "Location" field is set to the Volunteer Project, Location, "Street Address" field.
3. Activity custom field "Volunteer Role", based on the assigned role in the Volunteer assignment.

---

 * [VOL-312: Add more detail to volunteer activity when assigned \(TFC-94\)](https://issues.civicrm.org/jira/browse/VOL-312)

### DIFF
--- a/CRM/Volunteer/BAO/Assignment.php
+++ b/CRM/Volunteer/BAO/Assignment.php
@@ -247,10 +247,12 @@ class CRM_Volunteer_BAO_Assignment extends CRM_Volunteer_BAO_Activity {
       $defaults['campaign_id'] = '';
     }
 
+    // Taking volunteer_role_id outside of Action ADD scope, we have to update the role on moving the activity.
+    $defaults['volunteer_role_id'] = CRM_Utils_Array::value('role_id', $need);
     if ($op === CRM_Core_Action::ADD) {
-      $defaults['volunteer_role_id'] = CRM_Utils_Array::value('role_id', $need);
       $defaults['time_scheduled_minutes'] = CRM_Utils_Array::value('duration', $need);
-      $defaults['target_contact_id'] = CRM_Volunteer_BAO_Project::getContactsByRelationship($project->id, 'volunteer_beneficiary');
+      $defaults['target_contact_id'] = CRM_Volunteer_BAO_Project::getContactsByRelationship($project->id, 'volunteer_manager'); // Changed to volunteer_manager from volunteer_beneficiary.
+      $defaults['location'] = $project->getLocation(); //Getting location of the project to save in activity.
 
       // If the related entity doesn't provide a good default, use tomorrow.
       if (empty($params['activity_date_time'])) {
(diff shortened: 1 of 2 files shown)


CRM/Volunteer/BAO/Assignment.php
```diff
@@ -247,10 +247,12 @@ private static function setActivityDefaults(array $params) {
       $defaults['campaign_id'] = '';
     }
 
+    // Taking volunteer_role_id outside of Action ADD scope, we have to update the role on moving the activity.
+    $defaults['volunteer_role_id'] = CRM_Utils_Array::value('role_id', $need);
     if ($op === CRM_Core_Action::ADD) {
-      $defaults['volunteer_role_id'] = CRM_Utils_Array::value('role_id', $need);
       $defaults['time_scheduled_minutes'] = CRM_Utils_Array::value('duration', $need);
-      $defaults['target_contact_id'] = CRM_Volunteer_BAO_Project::getContactsByRelationship($project->id, 'volunteer_beneficiary');
+      $defaults['target_contact_id'] = CRM_Volunteer_BAO_Project::getContactsByRelationship($project->id, 'volunteer_manager'); // Changed to volunteer_manager from volunteer_beneficiary.
```
GinkgoFJG: @agileware, this change seems wrong to me. The target of the activity should be the beneficiary. Ultimately we want the activity to indicate that Joe Volunteer (activity assignee) was assigned by whomever (usually himself, but sometimes a staff member, if added in the backend -- anyway, the activity source or "added by" contact) to volunteer for Acme Charity (the activity target, aka the "with" contact).

A number of chaptered organizations are using CiviVolunteer, and this field allows them to distinguish between volunteers for Acme Charity New York vs. Acme Charity San Francisco.

CRM/Volunteer/BAO/Assignment.php
```diff
@@ -311,6 +313,12 @@ public static function createVolunteerActivity(array $params) {
       }
     }
 
+    // Saving the Volunteer role custom field for the activity.
```
GinkgoFJG: The following block should be unnecessary. The `foreach` loop immediately above this automagically converts `$params['volunteer_role_id']` to `$params["custom_$n"]`, and `$params` already takes `$defaults` into account a few lines before that. What kind of problems did you run into that you needed to add this code?

CRM/Volunteer/BAO/Project.php
```diff
@@ -169,6 +169,40 @@ public static function getContactsByRelationship($projectId, $relationshipType)
   }
 
   /**
+   * Gets location of a project.
```
GinkgoFJG: While I like the idea of passing location data into the activity, CiviVolunteer [already](https://github.com/civicrm/org.civicrm.volunteer/blob/master/api/v3/VolunteerProject.php#L226) has [so](https://github.com/civicrm/org.civicrm.volunteer/blob/5c0b4049f0391ede6d89d82f8988c561122225a6/ang/volunteer/Projects.js#L31) many [kludges](https://github.com/civicrm/org.civicrm.volunteer/blob/2da10302f9c0a6242bd75a8d1844c03aee62862c/ang/volunteer/Project.js#L196) for [dealing](https://github.com/civicrm/org.civicrm.volunteer/blob/59534288f37c96202434c57adc73d2fa5dc7ba55/volunteer.php#L580) with [locations](https://github.com/civicrm/org.civicrm.volunteer/blob/5c0b4049f0391ede6d89d82f8988c561122225a6/ang/volunteer/Projects.html#L89) that [I'm](https://github.com/civicrm/org.civicrm.volunteer/blob/5c0b4049f0391ede6d89d82f8988c561122225a6/ang/volunteer/Projects.js#L103) loathe to add another. And that's not to mention the kludges I have in customer implementations! I think it's high time we modify api.VolunteerProject.get to return more usable values so we don't have to keep doing string gymnastics.

How do you feel about working on that together instead of adding yet another locations hack?

jusfreeman:

PR #466, opened 2018-01-18
VOL-315: Fixed implicit assumption that group registration profiles will *always* have Activity fields.

GinkgoFJG:
* [VOL-315: Correct implicit assumption that group registration profiles will always have activity fields](https://issues.civicrm.org/jira/browse/VOL-315)


ginkgomzd:
:+1:

PR #467, opened 2018-03-14
Fixed logic error where Location form was hidden…

GinkgoFJG:
… when loc_block_id was the number 0, rather than the string '0'


GinkgoFJG:
This appears to introduce a bug (causing the location form to be visible always). Not only am I not merging this, I'm flagging it for removal downstream!

PR #468, opened 2018-03-17
Fix PHP 7.1 Error: Cannot pass parameter 1 by reference

seanmadsen:
Before this fix, visiting the "Manage events" page (when at least one event has an associated volunteer project, and the server is running PHP 7.1) would give the following error:

> Error: Cannot pass parameter 1 by reference in CRM_Volunteer_BAO_Project::retrieve() (line 478 of /home/sean/buildkit/build/ext/sites/default/files/civicrm/ext/org.civicrm.volunteer/CRM/Volunteer/BAO/Project.php).

This fix changes the call to `copyValues()` so that it passes a variable, making this code compatible with PHP 7.1 and above.

Here is a comparison of this behavior in different PHP versions: https://3v4l.org/b9Jea


GinkgoFJG:
Thanks, @seanmadsen.

PR #471, opened 2018-03-19
VOL-318: Addresses Backbone using CiviCRM's namespace.

GinkgoFJG:
* [VOL-318: Breakage caused by use of Backbone.noConflict\(\) introduced in CiviCRM 4.7.31](https://issues.civicrm.org/jira/browse/VOL-318)


GinkgoFJG:
Appears to address all issues manifesting as a result of the changes in CiviCRM 4.7.31. Needs further testing -- in particular I haven't tried to load CiviVolunteer in CiviEvent yet.

GinkgoFJG:
User QA passes, both in the "standalone" and CiviEvent contexts.

I'm pretty sure this breaks instances pre-4.7.31, but I'm not inclined to try to provide a patch that supports both releases.

PR #473, opened 2018-05-17
Whether passing through upgrade path or install path, ensure consistent naming of foreign key constraint

GinkgoFJG:
This problem was encountered while upgrading a very out-of-date CiviVolunteer to current.

Without this patch, the upgrade path results in a schema with an autogenerated name for one of its foreign key constraints. In the install path, the same constraint is explicitly provided a name, making it easy to reference in later queries. Later upgrade queries assume the constraint has the name supplied by the installer and will fail if the assumption proves false.


GinkgoFJG:
For reference: [the corresponding code in the install path](https://github.com/civicrm/org.civicrm.volunteer/blob/v4.7.31-2.3.0/sql/volunteer_install.sql#L9).

PR #479, opened 2018-09-10
Fixed compatibility with civicrm > 5.3

jaapjansma:
After upgrading to CiviCRM 5.4.1 we get the following error: 

**'Error in call to CustomField…_get : 'AdvMulti-Select' is not a valid option for field html_type'**

This PR fixes this error. This is caused by the fact that the AdvMulti-Select field is removed since CiviCRM 5.3


lkwetter:
Thank you Jaap for this fix.

jusfreeman:
This has been a problem since 31 May 2018, See https://github.com/civicrm/civicrm-core/pull/12238

jusfreeman:
@jaapjansma thanks for the PR / patch - tested and works. Good to merge IMHO.

PR #490, opened 2018-12-21
Add icon to top level menu entry

colemanw:
This makes the menu entry look nice.

Note: I couldn't think of a better icon than this. Even though it's already designated as signifying "groups" in civicrm I think it's ok to mean volunteers too.


ginkgomzd:
:+1:

PR #505, opened 2019-04-08
Change PHP keywords to comply with PSR2

mullikine:
Hi, I've changed a PHP keyword to comply with [this standard](https://www.php-fig.org/psr/psr-2/#25-keywords-and-truefalsenull) in PSR2. It's admittedly a small fix but I hope it helps!

See also,
https://github.com/civicrm/civicrm-wordpress/pull/149
https://github.com/civicrm/civicrm-drupal/pull/570

### DIFF
--- a/CRM/Volunteer/Angular.php
+++ b/CRM/Volunteer/Angular.php
@@ -2,7 +2,7 @@
 
 class CRM_Volunteer_Angular {
 
-  private static $loaded = FALSE;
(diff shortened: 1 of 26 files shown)


ginkgomzd:
Hello,
Thanks for the PR, but I'm going to close it without merging because even though there are not explicit guidelines for style for CiviCRM extensions, the default standard is the Drupal standards.

Cheers

PR #517, opened 2019-08-01
Ensure that Volunteer Report instance can only be accessed by people …

seamuslee001:
…with access to CiviReport

At present if someone was to go to the public url civicrm/report/list they would potentially be able to create copies, run the report etc. This ensures that you have to at least access CiviReport to do anything with the report

ping @jkingsnorth @totten @eileenmcnaughton @ginkgomzd

### DIFF
--- a/CRM/Volunteer/Form/VolunteerReport.mgd.php
+++ b/CRM/Volunteer/Form/VolunteerReport.mgd.php
@@ -32,6 +32,7 @@ return array(
       'title' => $labelVolunteerReport,
       'description' => $labelVolunteerReport,
       'report_id' => 'volunteer',
+      'permissions' => 'access CiviReport',
       'form_values' => serialize(array(
         'fields' => array(
           'contact_assignee' => '1',


seamuslee001:
also ping @mlutfy

ginkgomzd:
Thanks @seamuslee001 !
But, are you sure this is needed?
I'm spot checking on one instance I don't see that a user without Access CiviReports can see anything at that URL. 
I don't get access denied, but it says, "No reports have been created.  Contact your site administrator for help creating reports.", which is not true, so seems fine without this change.

seamuslee001:
@ginkgomzd try https://civicrm.org/civicrm/report/list

ginkgomzd:
huh, that's interesting... I just tried a bunch of random civicrm sites... and none of them are displaying that. Was there a change in the permission checks in core?

I'm not against the change, just curious what is going on.

seamuslee001:
I'm not sure maybe @mlutfy can check and see what is going on perhaps. I haven't seen any changes to this https://github.com/civicrm/civicrm-core/blob/master/CRM/Report/Page/InstanceList.php#L164 or https://github.com/civicrm/civicrm-core/blob/master/CRM/Report/Utils/Report.php#L346 to suggest that there have been changes in the core code but not sure

mlutfy:
It's odd: the `civicrm_report_instance` table had empty values for the permissions/roles. I re-saved the report instance, and then it was fine (i.e. saved in DB).

Issue #518, opened 2019-09-03
Image / file upload issue on volunteer public form


adamt-devapp:
Hi,

I am just wondering if file uploading (image) is supported in CiviVolunteer as I am having an issue where attaching an image to a volunteer signup form is not being captured / displayed on the view contact details page.

Is this possibly a bug or just not supported by CiviVolunteer?

For clarity on what I have done so far and the results I am experiencing.

We want to add a file upload option on volunteer sign up page that allows the volunteer to add a picture of themself.

So I created a new custom field (file upload) and assigned it to a volunteer application form profile with expected results.
i.e: Field is displayed on public form and also available on the contact record.

However, once the form has been submitted and the new contact created the image is not available on the contact details page.

I assumed the image would be rendered next to the label or possibly a link to view elsewhere.

When editing the contact details from within Civi I am able to upload an image and then upon saving it is rendered on screen.
So it seems that the file upload is working but just not via the volunteer form.

I have attached screenshots to detail what is going on.

Thanks.

Regards,
Adam.T
![Upload_Image_Missing_Contacts_View](https://user-images.githubusercontent.com/54293902/64152125-17f0f280-ce6b-11e9-8f1c-695b0fe11e45.png)
![File_Upload_Volunteer_Form](https://user-images.githubusercontent.com/54293902/64152158-2ccd8600-ce6b-11e9-95e9-66573dd36ea8.png)
![contact_profile_upload_image](https://user-images.githubusercontent.com/54293902/64152187-3b1ba200-ce6b-11e9-8fe9-d398b50c9d9e.png)
![Upload_image_from_view_contact](https://user-images.githubusercontent.com/54293902/64152197-41118300-ce6b-11e9-9078-fa118bd881e0.png)

ginkgomzd:
Hi @adamt-devapp 

That is curious that the file upload is working on other forms, but not he volunteer sign-up. I'm not sure that you have tested it with another profile though.

The volunteer sign-up form is not a feature of CiviVolunteer. It is just another profile.

The first thing I think of is permissions on the destination directory.  But since it is working sometimes, my theory doesn't sound very good. You'll have to look for more information. Maybe you can find an error message in your browser developer console.

You should try requesting help on the CiviCRM Stack Exchange community.

Cheers

PR #519, opened 2019-09-05
Fixed Image upload field processing on volunteer signup form

pradpnayak:
For https://github.com/civicrm/org.civicrm.volunteer/issues/518 and https://civicrm.stackexchange.com/questions/31933/image-file-upload-issue-on-civivolunteer-public-form


ginkgomzd:
I replied on the SE thread.
This PR looks fine, but I'd like to understand what is going on.
Right now, this is looking to be caused by a change in core, so I'd like to know what version the change was introduced. In particular, is this fix backwards compatible? I can go check that right now...

PR #523, opened 2019-09-24
F/display edit custom data

davidjosephhayes:
Took a shot at getting custom data editable on Volunteer Projects.

### DIFF
--- a/CRM/Volunteer/BAO/Project.php
+++ b/CRM/Volunteer/BAO/Project.php
@@ -36,6 +36,13 @@
 class CRM_Volunteer_BAO_Project extends CRM_Volunteer_DAO_Project {
 
   /**
+   * Static field for all the case information that we can potentially export.
+   *
+   * @var array
+   */
+  public static $_exportableFields = NULL;
+
+  /**
    * Array of attributes on the related entity, translated to a common vocabulary.
    *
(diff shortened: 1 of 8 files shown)


ginkgomzd:
Hi @davidjosephhayes ,
Thanks for submitting this!
I'm in a rush right now, so I'm not sure I'm groking what this PR is all about and what the ramifications of it might be.

What is your situation with this? Are you just posting in the hope it might help someone in the future? If you plan to work on this more, I think we should talk because some work we have done in custom forks is likely to be relevant. If you are able to work on this in the future, it would be great to get a unified vision for it.

Shoot me your contact info? michael@ginkgostreet.com
I can't meet this week, but would be happy to do a screenshare next wek.

Cheers!

davidjosephhayes:
Sent you an email.

I will be working more on this in the coming few weeks as the project progresses.

We are using https://contactform7.com/ with https://wordpress.org/plugins/contact-form-7-civicrm-integration/ to push volunteer projects from staff into the system via the api. Much of what we collect is going to be custom data. I adapted much of the code from https://github.com/civicrm/org.civicrm.civicase.

I ran through a few of the forks, but did not find what I was looking for. Which one should I take a look at?

PR #532, opened 2020-02-23
Fix notice against 5.24.0 due to core function signature change

eileenmcnaughton:
The signature for copyValues in 5.24.0 no longer has serialize as a second param. This removes
& increments the version so only sites on 5.24.0+ should get the change (This is the strategy I
use in all my extensions - ie specifying which civi version each release is tested on.)

### DIFF
--- a/CRM/Volunteer/BAO/Project.php
+++ b/CRM/Volunteer/BAO/Project.php
@@ -632,7 +632,7 @@ class CRM_Volunteer_BAO_Project extends CRM_Volunteer_DAO_Project {
    * @deprecated since version 4.7.21-2.3.0
    *   Internal core methods should not be extended by third-party code.
    */
-  public function copyValues(&$params, $serializeArrays = FALSE) {
+  public function copyValues($params) {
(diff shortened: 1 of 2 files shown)


eileenmcnaughton:
Screen Shot 2020-02-24 at 11 15 34 AM

ginkgomzd:
Thanks @eileenmcnaughton ,
I'm taking a look at refactoring this because it looks like the only reason this override was created was to default one field.

I have more free time now, so I will see if I can finally get that release out. I recall that I found a bug in the rc. I will try to take the shortest path to a release.

eileenmcnaughton:
@ginkgomzd would be great to see that release out

JoeMurray:
+:+1: about getting a release out. There seem to have been a few rc branches created but no release in a long time.

ginkgomzd:
superseded by #536  

particulary: https://github.com/civicrm/org.civicrm.volunteer/pull/536/commits/b1ab67269fb992ce96a4758b29ab86131abd53d8

PR #537, opened 2020-03-16
Resolve #533 by ensuring entity details are correctly loaded for use …

### LINKED ISSUE #533: Unable to edit profile from Event Volunteer configuration tab
After installing CiviVolunteer a new tab is added to Event Configuration which allows you to define a volunteer opportunity for a specific event. Down the bottom of the page you can select a profile to be used when people register their interest in volunteering. There is an edit button next so you may be able to edit the profile however when you click it, it doesn't load and just stays looking like your trying to load and an error appears in the console about failing to load entity contact_1

seamuslee001:
…by the interactive profile editor

ping @ginkgomzd

### DIFF
--- a/CRM/Volunteer/Angular/Tab/Event.php
+++ b/CRM/Volunteer/Angular/Tab/Event.php
@@ -44,6 +44,9 @@ class CRM_Volunteer_Angular_Tab_Event extends CRM_Core_Page {
     $event = $project->getEntityAttributes();
     $entityTitle = $event['title'];
 
+    CRM_UF_Page_ProfileEditor::registerProfileScripts();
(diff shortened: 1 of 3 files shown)


ginkgomzd:
Thanks @seamuslee001 !
However this breaks the manage project form. I think I can make them work together.

$usedFor is undefined. Passing an empty array instead seems to work. Do you think the widget requires this?

seamuslee001:
@ginkgomzd I'm not 100% sure but I suspect it might be, i was comparing this to the online registration form html to see what the diffs were between the profile links there and on the volunteer tab

seamuslee001:
@ginkgomzd sorry michael for the lag on this but I think this is now fixed for you

PR #550, opened 2020-11-18
ensure searches are inclusive of end date

jmcclelland:
If you search for 2020-11-19 - 2020-11-19, it should show
opportunities on 2020-11-19. With this patch it does, without
this patch it does not.


ginkgomzd:
Thanks Jaime!

PR #551, opened 2021-02-04
Regenerate Civix Generated files

seamuslee001:
ping @eileenmcnaughton @colemanw

### DIFF
--- a/CRM/Volunteer/Upgrader/Base.php
+++ b/CRM/Volunteer/Upgrader/Base.php
@@ -1,6 +1,7 @@
 <?php
 
 // AUTO-GENERATED FILE -- Civix may overwrite any changes made to this file
+use CRM_Volunteer_ExtensionUtil as E;
 
 /**
  * Base class which provides helpers to execute upgrade logic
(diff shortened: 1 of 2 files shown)


seamuslee001:
ping @MikeyMJCO as well

eileenmcnaughton:
@MikeyMJCO @ginkgomzd I feel like I tried to broker a co-maintainer relationship here but I think I failed at the post - can I propose again that @MikeyMJCO is added as co-maintainer & we tag a release on this?

ginkgomzd:
Mikey and I agreed to work on maintaining the extension and left off seeing who would have time to test first.

Let's focus on the issue at hand, I have little care for who has tagging rights on the repo.

I think if there is anyone, not requesting, but vouching for a release, then it should be tagged as such.

eileenmcnaughton:
I'm pretty sure I can vouch for the master branch being more reliable that the current release based on all the things that have changed in master to adapt to civi changes that are not released

We should probably decide whether a release is going to happen in x period of time & if not consider removing this from buildkit / demo sites

ginkgomzd:
Okay, as long as we are vouching based on commits...

I reviewed the two release candidates and remembered that one was a conservative release and the other is known to have bugs.

Master already superseded the conservative release and those commits have more assurance than most.

I pushed the tag for v2.4.0.

eileenmcnaughton:
thanks @ginkgomzd that's a relief since I've spent a bit of time telling people to download from github rather than use the tagged release

PR #555, opened 2021-03-31
Expose field “Address Name” in volunteer project location

yashodha:
Expose field “Address Name” in volunteer project location. This is in keeping with other location displays like contacts, events, etc.

volunteer_addr


ginkgomzd:
Thanks!

PR #556, opened 2021-04-01
add Volunteer project description to activity details for context

yashodha:
add Volunteer project description to activity details for context


ginkgomzd:
Hi Yashoda!

Thanks for this, but I'm going to decline to merge this because I'm unsure how universally useful this is.
I'm aware of some instances that use very lengthy project descriptions. The Activity subject already represents the Project Title, so I don't think there is a complete lack of context. I think a better improvement might be to provide a link to the project.

If you want to make a case for this, I'm always open to discussion and curious how users are using volunteer.

Cheers

PR #559, opened 2021-04-09
Change the unit of the duration of a volunteer opportunity

yashodha:
looks cumbersome to enter in minutes

### DIFF
--- a/CRM/Volunteer/Form/Log.php
+++ b/CRM/Volunteer/Form/Log.php
@@ -203,7 +203,7 @@ class CRM_Volunteer_Form_Log extends CRM_Core_Form {
       if (!$duration) {
         $errors["field[$key][actual_duration]"] =
           ts('Please enter the actual duration volunteered.', array('domain' => 'org.civicrm.volunteer'));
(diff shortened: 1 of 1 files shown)


ginkgomzd:
I think this would need an upgrade step for existing systems.

I believe this is a legacy rooted in how activities represent time, but I'd have to dig for what the precedent actually was.

I do know that the position has been that this is best handled by formatting in displays.

yashodha:
@ginkgomzd I agree this was based on how we always show activity duration in minutes. But this is a just display thing where we show the volunteer time in hours ( since it could be cumbersome to convert to minutes when you are logging 5/6 hours). I don't think any changes to the upgrade are warranted as it shouldn't affect how data is stored.

ginkgomzd:
ah, I didn't look carefully at what you were doing here.

Did you try to break it entering fractional hours?
How about converting from 20 minutes, 13 minutes...

I'll give this a proper look this week or next.

Cheers

yashodha:
@ginkgomzd what do you think about this? Can you please take a look at this PR? Thanks!

PR #563, opened 2021-05-02
Fix uninstall and install routines.

MikeyMJCO:


MikeyMJCO:
Tested and working on Drupal 8/9 and Wordpress. 

Separate issue where the uninstall steps in `upgrader.php` aren't being called on Uninstall properly!

ginkgomzd:
Looks like this arose with this merge:
https://github.com/civicrm/org.civicrm.volunteer/commit/e03d9f5717ff1e7365d0b1685b9ef7ac8f11354d#diff-d2e789e15337a762dfca1a44d240ff5c43a76976b8bd9eb61344c708be154fe8

Can you look at just making these upgrade step calls instead?
```php
$this->schemaUpgrade20();
$this->addNeedEndDate();
$this->installNeedMetaDateFields();
```

ginkgomzd:
I've suspected there were two bugs... this looks like the one that was less-well-reported.

It would be nice to find the proof that these bugs are one and the same.
i.e., one bug is: schema/entities already exist, the other is, schema/entities were not created.

MikeyMJCO:
So I've been able to confirm that some parts haven't been being created on Install, when you say "look at making these upgrade step calls" you mean rather than calling `$this->installProjectRelationships();` call:

```php
$this->schemaUpgrade20();
$this->addNeedEndDate();
$this->installNeedMetaDateFields();
```

MikeyMJCO:
The "Already Exist" bug is different - for some reason the steps in `Upgrader.php` -> uninstall() aren't being called on uninstall.

MikeyMJCO:
@ginkgomzd this now has a pretty major refactor of the uninstall routines (based on the GDPR extension's implementation) 

I have one question remaining for you - do you think we should remove the custom activity statuses added? I've gone with "no" since they don't appear to cause a conflict and removing them changes the activity records in the database - which I'm not keen on since I don't think there's a case for removing those activities on uninstall.

ginkgomzd:
> So I've been able to confirm that some parts haven't been being created on Install, when you say "look at making these upgrade step calls" you mean rather than calling `$this->installProjectRelationships();` call:
> 
> ```
> $this->schemaUpgrade20();
> $this->addNeedEndDate();
> $this->installNeedMetaDateFields();
> ```
> 
> Those 3 calls are in the `install()` step already - the missing one was `installProjectRelationships()`

how do you figure?... I don't see those calls, except in upgrade steps.

MikeyMJCO:
> > So I've been able to confirm that some parts haven't been being created on Install, when you say "look at making these upgrade step calls" you mean rather than calling `$this->installProjectRelationships();` call:
> > ```
> > $this->schemaUpgrade20();
> > $this->addNeedEndDate();
> > $this->installNeedMetaDateFields();
> > ```
> > 
> > 
> > Those 3 calls are in the `install()` step already - the missing one was `installProjectRelationships()`
> 
> how do you figure?... I don't see those calls, except in upgrade steps.

You're right - not sure what I was seeing when I looked yesterday :-) So add those in addition to `installProjectRelationships()`?

ginkgomzd:
> 
> You're right - not sure what I was seeing when I looked yesterday :-) So add those in addition to `installProjectRelationships()`?

;-)

yes, but need to reconcile that upgrade20 does:
```php
$this->installProjectRelationships();
$this->executeSqlFile('sql/volunteer_upgrade_2.0.sql');
```
... so, maybe just the three calls so we keep it DRY

MikeyMJCO:
Is calling the upgrade SQL going to cause a problem on install - is it better to do `installProjectRelationships();` instead of `upgrade20()`

ginkgomzd:
... Activity Statuses... I agree, don't delete them.

ginkgomzd:
> Is calling the upgrade SQL going to cause a problem on install - is it better to do `installProjectRelationships();` instead of `upgrade20()`

looking it over... I think refactoring-out the upgrade SQL is a little risky. I think not calling `installProjectRelationships()` explicitly is better for now.

MikeyMJCO:
Done - made those changes!

ginkgomzd:
@MikeyMJCO, confirming you think this is ready to merge? No longer WIP?

MikeyMJCO:
Yeah, I am!

PR #564, opened 2021-05-08
Fix Install routine and regenerate DAOs

seamuslee001:
ping @MikeyMJCO @ginkgomzd 

This fixes errors installing doing the following which is from https://github.com/civicrm/org.civicrm.volunteer/blob/1a0edeab91bd2edd3cd63a0c602fc28c0aeb2c3c/sql/volunteer_upgrade_2.0.sql#L30 

```
+++ drush8 -y cvapi extension.install key=org.civicrm.volunteer debug=1
Cannot execute INSERT INTO `civicrm_volunteer_project_contact` (`project_id`, `contact_id`, `relationship_type_id`)
SELECT `id`, `target_contact_id`, @beneficiary_reltype_id
FROM `civicrm_volunteer_project`
WHERE `target_contact_id` IS NOT NULL: DB Error: no such fieldDrush command terminated abnormally due to an unrecoverable error.       [error]
+ set +x
```


MikeyMJCO (approved):

ginkgomzd:
Thanks @seamuslee001 !

PR #571, opened 2021-06-07
Remove unused clientside variable

colemanw:
This variable `currentContactId` is pushed to the client side but never accessed.
FYI if it were needed, core already makes it available as `CRM.config.cid`.


ginkgomzd:
This variable probably shouldn't be used, but I did confirm there are custom implementations that depend on it. It would be proper to deprecate this in one release before removing.

colemanw:
@ginkgomzd what other extensions use it? I can send them up a patch.

PR #578, opened 2021-09-17
Update campaign field with component, html, pseudoconstant attrs

colemanw:
Updates the campaign field to be consistent with the ones in core, and regenerates code.

This is a noisy PR because of the whitespace changes in autogenerated files, but the only functional change is to add metadata to the campaign field.


colemanw:
Thanks @ginkgomzd. I've been holding off on other refactoring work because you said you have uncommitted changes staged - is that still the case?

PR #579, opened 2021-10-22
copy over duration to Volunteer activity

yashodha:
copy over duration to Volunteer activity


ginkgomzd:
okay, I see that we do have a custom field, time_completed_minutes, that holds logged time, not the duration field.

PR #589, opened 2023-04-11
Replace deprecated CRM_Core_OptionGroup::getValue()

MegaphoneJon:
Civi 5.60 removes `CRM_Core_OptionGroup::getValue()`. This updates the code to the non-deprecated equivalent.

Note that I didn't get every single one in the tests - I'm not convinced any of my clients are using this, so I put in the effort needed to make sure crashes didn't happen.


ginkgomzd:
Thanks @MegaphoneJon !

PR #590, opened 2023-05-15
Prevent hard-failure on latest CiviCRM

eileenmcnaughton:
There is still call in the extension to a function that has been progressively deprecated since 4.7 - which causes a fatal error on the ManageFees page if CiviVolunteer is enabled (in the js)

There are still some calls in the tests but they can't have been run in years as they would have failed on the deprecation warning

### DIFF
--- a/CRM/Volunteer/Form/VolunteerReport.mgd.php
+++ b/CRM/Volunteer/Form/VolunteerReport.mgd.php
@@ -1,7 +1,7 @@
 <?php
 
 /**
- * This file declares managed database records of type "ReportTemplate" and
+ * This file declares managed database records of type "OptionValue" and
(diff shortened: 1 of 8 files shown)


CRM/Volunteer/Form/VolunteerReport.mgd.php
```diff
@@ -1,7 +1,7 @@
 <?php
 
 /**
- * This file declares managed database records of type "ReportTemplate" and
+ * This file declares managed database records of type "OptionValue" and
```
eileenmcnaughton, line 4: I included this by accident - but it is a change being made across the ReportTemplates as we switch to apiv4

eileenmcnaughton:
I merged Jons

PR #591, opened 2023-05-22
Update info.xml - tag release cos this hard-fails on 5.60+ without a new release

eileenmcnaughton:
@ginkgomzd @homotechsual - we really need to tag this to avoid a hard fail on a very old long deprecated function that was removed. I incremented the version to 5.60 since earlier Civi versions are 'ok' at least on this issue so incrementing the version is more conservative

 There is another problem that persists even with this on saving events on the latest civi - but I would do a second release if we figure that out rather than delay getting this important fix out

### DIFF
--- a/info.xml
+++ b/info.xml
@@ -8,10 +8,10 @@
     <author>Ginkgo Street Labs</author>
     <email>ginkgomzd@fastmail.com</email>
   </maintainer>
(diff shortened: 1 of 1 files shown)


ginkgomzd:
Thanks @eileenmcnaughton

PR #599, opened 2024-03-13
Fix deprecation warnings for permissions array in 5.71

cviebrock:
As of CiviCRM 5.71, these warnings appear:

```
User deprecated function: Permission 'register to volunteer' should be declared with 'label' and 'description' keys. See https://docs.civicrm.org/dev/en/latest/hooks/hook_civicrm_permission/ Caller: CRM_Core_Permission::assembleBasicPermissions in CRM_Core_Error::deprecatedWarning() (line 1129 of /var/www/html/vendor/civicrm/civicrm-core/CRM/Core/Error.php).

User deprecated function: Permission 'log own hours' should be declared with 'label' and 'description' keys. See https://docs.civicrm.org/dev/en/latest/hooks/hook_civicrm_permission/ Caller: CRM_Core_Permission::assembleBasicPermissions in CRM_Core_Error::deprecatedWarning() (line 1129 of /var/www/html/vendor/civicrm/civicrm-core/CRM/Core/Error.php).

...
```

... along with similar warnings about the rest of the permissions.

I've not done any work with Civi extensions, but looking into the URL provided (https://docs.civicrm.org/dev/en/latest/hooks/hook_civicrm_permission/), I believe the issue is related to how the permissions are configured in `CRM/Volunteer/Permission.php`:

```php
    return array(
      'register to volunteer' => array(
        $prefix . ts('register to volunteer', array('domain' => 'org.civicrm.volunteer')),
        ts('Access public-facing volunteer opportunity listings and registration forms', array('domain' => 'org.civicrm.volunteer')),
      ),
      'log own hours' => array(
        $prefix . ts('log own hours', array('domain' => 'org.civicrm.volunteer')),
        ts('Access forms to self-report performed volunteer hours', array('domain' => 'org.civicrm.volunteer')),
      ),
      ...
```

This array should, I think, be converted to:

```php
    return array(
      'register to volunteer' => array(
        'label' => $prefix . ts('register to volunteer', array('domain' => 'org.civicrm.volunteer')),
        'description' => ts('Access public-facing volunteer opportunity listings and registration forms', array('domain' => 'org.civicrm.volunteer')),
      ),
      'log own hours' => array(
        'label' => $prefix . ts('log own hours', array('domain' => 'org.civicrm.volunteer')),
        'description' => ts('Access forms to self-report performed volunteer hours', array('domain' => 'org.civicrm.volunteer')),
      ),
      ...
```

etc..  i.e. turn the non-associative array into an associative one with `label` and `description` keys.

This PR does that. 😄 (It also reduces the constant definition of `array('domain' => 'org.civicrm.volunteer')` and just does that once, similar to how the `$prefix` variable is used.)

### DIFF
--- a/CRM/Volunteer/Permission.php
+++ b/CRM/Volunteer/Permission.php
@@ -11,43 +11,44 @@ class CRM_Volunteer_Permission {
    * @return array Keyed by machine names with human-readable labels for values
    */
   public static function getVolunteerPermissions() {
-    $prefix = ts('CiviVolunteer', array('domain' => 'org.civicrm.volunteer')) . ': ';
+    $domain = array('domain' => 'org.civicrm.volunteer');
+    $prefix = ts('CiviVolunteer', $domain) . ': ';
     return array(
       'register to volunteer' => array(
-        $prefix . ts('register to volunteer', array('domain' => 'org.civicrm.volunteer')),
-        ts('Access public-facing volunteer opportunity listings and registration forms', array('domain' => 'org.civicrm.volunteer')),
+        'label' => $prefix . ts('register to volunteer', $domain),
+        'description' => ts('Access public-facing volunteer opportunity listings and registration forms', $domain),
       ),
       'log own hours' => array(
-        $prefix . ts('log own hours', array('domain' => 'org.civicrm.volunteer')),
-        ts('Access forms to self-report performed volunteer hours', array('domain' => 'org.civicrm.volunteer')),
+        'label' => $prefix . ts('log own hours', $domain),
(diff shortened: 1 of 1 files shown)


ginkgomzd:
Thanks Colin!

cviebrock:
Thanks @ginkgomzd ... for planning purposes on our end, is there an ETA for a new version to be tagged?

ginkgomzd:
No. 

For installs I maintain, I do not "plan" when an instance diverges from official releases. I typically create patch files and document them by putting them in a safe place, ideally with reference to an issue or PR so it will be clear when upstream obviates the patch.

PR #602, opened 2024-05-10
Cleanup: Avoid deprecated internal variable $_tableName

colemanw:
Cleans up code that was using a deprecated-soon-to-be-removed internal property.


ginkgomzd:
obviated by #608

PR #604, opened 2024-05-21
Avoid errors when end_time is not specified

jmcclelland:
I'm not sure if this code ever worked. The old code was running date_add on the start_time as a string instead of an object.

I re-organized the code a little to make it more readable.


ginkgomzd:
Thanks Jamie!
Being in the Roster report, it very likely could have never worked and gone unnoticed.

Issue #577, opened 2021-09-09
Volunteer Opportunities on WordPress permissions error


kelizoliva:
Since the most recent CiviCRM Security Release (5.40.3), though possibly earlier, we have been experiencing the follownig error for a Civi-WordPress implementation:

```
Sep 09 13:30:00  [error]
$Fatal Error Details = array:3 [
  "message" => "You do not have permission to access this page."
  "code" => null
  "exception" => CRM_Core_Exception {#4059
    -errorData: array:1 [
      "error_code" => 0
    ]
    #cause: null
    -_trace: null
    #message: "You do not have permission to access this page."
    #code: 0
    #file: "/var/www/test/web/wp-content/plugins/civicrm/civicrm/CRM/Utils/System/WordPress.php"
    #line: 605
    trace: {
      /var/www/test/web/wp-content/plugins/civicrm/civicrm/CRM/Utils/System/WordPress.php:605 {
        ›   status_header(403);
        ›   throw new CRM_Core_Exception(ts('You do not have permission to access this page.'));
        › }
      }
      /var/www/test/web/wp-content/plugins/civicrm/civicrm/CRM/Utils/System.php:60 { …}
      /var/www/test/web/wp-content/plugins/civicrm/civicrm/CRM/Core/Invoke.php:233 { …}
      /var/www/test/web/wp-content/plugins/civicrm/civicrm/CRM/Core/Invoke.php:69 { …}
      /var/www/test/web/wp-content/plugins/civicrm/civicrm/CRM/Core/Invoke.php:36 { …}
      /var/www/test/web/wp-content/plugins/civicrm/civicrm.php:1178 { …}
      /var/www/test/web/wp-includes/class-wp-hook.php:303 { …}
      /var/www/test/web/wp-includes/class-wp-hook.php:327 { …}
      /var/www/test/web/wp-includes/plugin.php:518 { …}
      /var/www/test/web/wp-includes/class-wp.php:763 { …}
      /var/www/test/web/wp-includes/functions.php:1291 { …}
      /var/www/test/web/wp-blog-header.php:16 { …}
      /var/www/test/web/index.php:17 { …}
    }
  }
]
```
This is when attempting to, as an anonymous user, access `crm/?page=CiviCRM&q=civicrm/vol/#/volunteer/opportunities`

We/the client have not made any changes to permissions since this was last working. We have AJAX API and CiviVolunteer: register to volunteer permissions are granted on anonymous users.

[...]

PR #607, opened 2024-06-05
org.civicrm.volunteer#577 Resolving incorrect backend url generation

shaneonabike:
As indicated in #577 this fix resolves incorrect generated `volunteer/signup` URLs that are generated as backend URLs rather than frontend ones.


ginkgomzd:
without proper rigor, merging this optimistically. If a bug pops up, I hope you are good to handle that as well. ;-)

Thanks for the fix! I think this has been out there for awhile.

PR #617, opened 2024-10-24
Fix CiviCRM 5.70+ and smarty3+ compatibility

mlutfy:
I tried fiddling with the mixins to see if it made a difference (and updated the min-required to 5.71, for the latest smarty mixins), but I don't think there is an easy fix for making the `templates` directory available during the installation.

So I resorted to using the `postInstall ` hook for everything.

Also made some minor changes to the settings form, to fix some PHP notices (qfkey and other form elements were causing undefined array warnings).


stesi561:
Hmm. I'm still seeing a             [error_message] => Unable to load 'file:volunteer-customdata.xml.tpl'
when trying to install. Not I've also got https://github.com/civicrm/org.civicrm.volunteer/pull/618 on.

stesi561:
apologies. When digging into what was going on I realised that the patch failed to apply completely due a permissions error part way through. Redownloading master, reapplying this and then applying #618 and the install runs.

Issue #609, opened 2024-08-09
Unable to install on dmaster.demo.civicrm.org


petednz:
after enabling Angular Profiles and then trying to install this extension i get "The website encountered an unexpected error. Please try again later."

I enabled the Log Viewer and retried and again got above error but nothing showing in View Log.

Maybe something in Drupal logs but can't access them

mlutfy:
Here is a fix: https://github.com/civicrm/org.civicrm.volunteer/pull/617

I will close this issue in favour of #615

stesi561:
Will need this as a fix as well I believe. https://github.com/civicrm/org.civicrm.volunteer/pull/618

PR #618, opened 2024-10-24
Refs: #609 - account for new required field

stesi561:
Before: 
Attempt to install extension and get a mysterious DB Error. Extension Fails to install.
After: 
Attempt to install extension and get a different error - addressed in https://github.com/civicrm/org.civicrm.volunteer/pull/617

Note probably worth someone else r-runing this as nobody has touched these files for, a while, and changing without tests makes me nervous!


ginkgomzd:
Thanks @stesi561 !

PR #639, opened 2026-08-25
Extensive re-design and refactor

jackrabbithanna:
This PR contains:

- Updates for CiviCRM 6.16+, PHP8.4, and Smarty 5
- Removing dependency on angular profiles
- API3 to API4 conversion
- Updated / user friendly design that gels with Riverlea

This is a prototype, we're excited about the potential here, but still considering architecture updates. 

1. In particular integrating with CiviCRM's test infrastructure / buildkit better should be done
2. We're considering removing profiles completely and using Afform / Search Kit instead (we do use it for the new Hours Report) 
3. We're evaluating the JS architecture

Please let us know if you're interested in merging, or need changes to merge. We're happy to work together to make a new release!

Skvare is open and appreciates discussion and PRs on our fork based on community feedback to improve the extension further regardless of decision to merge by this repo's maintainer.


ang/volunteer/VolOppsCtrl.js
```diff
@@ -1,237 +1,433 @@
-(function (angular, $, _) {
+(function(angular, _, $) {
```
colemanw, line 1: Reordering params for no reason is a very LLM thing to do lol. But we are in the process of deprecating lo-dash so it'd be best not to use it at all.

colemanw:
@jackrabbithanna I started in, but the sheer volume of code makes this PR practically unreviewable. Maybe your intent was for user testing to check the "vibe" of the changes rather than developer review. But to achieve the latter I would recommend breaking it into *much* smaller commits, e.g.
- Update api3 to api4 in class X
- Update api3 to api4 in class Y
- Bump Smarty to v5
- Run `civix upgrade`
- Remove AngularProfiles dependency
- etc...

shaneonabike:
I still feel like we could get a slimer product if we just rewrote the entire thing using Searchkit and FormBuilder. Then maybe port over some of the additional functionality? Also doing so would mean that we could make this extension even more containerable (meaning that people could turn on X,Y,or Z elements within it depending on their needs).

I have a client who wants this done, but it didn't get in for this year. We shall see for the next year. Or maybe it could be an awesome thing to start at a sprint.

mlutfy:
I did some testing, although not exhaustive, but it works and it looks good. I agree with Coleman, it's a huge PR, difficult to review.  I agree with Shane about SearchKit, but some bits are difficult to do with SearchKit currently. I like that this PR removes the dependency on the deprecated AngularProfiles. If split in separate PRs, this could provide an upgrade path forward for existing users.

Some stats, out of the 274 files changed:

- 102 are for tests, tools, docs, dot-files, mixins, README
- 45 are for adding APIv4
- 10 are changes to APIv3
- 12 are for Backbone removals 

That still leaves around 100 files for user-facing changes (I used `git diff --name-status master..redesign-refactor` to review file changes).

Side-quest: There are still some broken icons. Best as a separate PR, but we can search-replace:

`ui-icon-(circle-check)` => `fa fa-$1`  ex: `fa fa-circle-check`

```
$ grep -r 'ui-icon'
ang/volunteer/Project.html:            <span class="crm-button-icon ui-icon-comment"> </span>
ang/volunteer/Project.html:        <span class="crm-button-icon ui-icon-triangle-1-e"> </span>
ang/volunteer/Project.html:        <span class="crm-button-icon ui-icon-circle-check"> </span>
ang/volunteer/Project.html:        <span class="crm-button-icon ui-icon-close"> </span>
js/commendation.js:        $('[data-identifier="_qf_Commendation_submit_delete"] .ui-icon' )
js/commendation.js:          .removeClass('ui-icon-check').addClass('ui-icon-trash');
```

jackrabbithanna:
@shaneonabike I almost replaced profiles for Form Builder, and some of the lists for Search kits, but thought that would be too much change for the first update. I agree it should be done, and if we get traction to get this committed, then we'll definitely do it in the future. I thought this was a good compromise to get started.

